PR #1, opened 2024-09-09
Constructors

alfiopuglisi:
Added main simulation object simul.py
Starting to move the functionality in factory.py inside each class


FabioRossiArcetri (commented):
Look ok tp me.

PR #4, opened 2024-09-13
Single prop

alfiopuglisi:
Added inputs/outputs to most objects used in main_scao.py

Propagation creates outputs automatically based on the list of input sources


alfiopuglisi:
I had to add another parameter type "xxx_dict", because propagation needs to know both the source name and the source reference in order to create the outputs.

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #5, opened 2024-09-17
Merge branch

FabioRossiArcetri:


alfiopuglisi:
Something happened to this branch, there should not be the full list of commit from the main one.
If I try merging manually from a local copy I get this error: "fatal: refusing to merge unrelated histories"
Probably we need to somehow just select the last five commits and apply those instead of merging

alfiopuglisi:
I was able to cherry-pick the branch commits and pushed them to main.
Everything is updated, this pull request can be deleted.
There is one bug left, phasescreen generation must be done with CPU otherwise the input turbulence is too big.

PR #19, opened 2024-09-30
Bugfixes

alfiopuglisi:
Fixes for various bugs found mixing GPU and CPU


pyssata/base_time_obj.py
FabioRossiArcetri, line 1: This should be >= than 0

FabioRossiArcetri:
also note the conflicts

pyssata/base_time_obj.py
alfiopuglisi: Fixed, thanks

FabioRossiArcetri (approved):
ok now

PR #24, opened 2024-10-02
(GA) missing int in indices and bugs in read method

guidoagapito:


pyssata/data_objects/iirfilter.py
```diff
@@ -245,13 +247,16 @@ def save(self, filename, hdr=None):
         hdul.writeto(filename, overwrite=True)
 
     def read(self, filename, hdr=None, exten=0):
```
alfiopuglisi: We can wait before working on the read/save methods, we still have to define an overall strategy for them

guidoagapito: I went back to old read method.
I kept other changes and updated discrete_delay_tf

PR #26, opened 2024-10-07
FFT shape

alfiopuglisi:
Changed FFT shape from np x np x N to N x np x np

SCAO case from 175 to 210 Hz
ELT case from 5 to 30 Hz


FabioRossiArcetri (approved):
OK

PR #30, opened 2024-10-08
Multiple yml files for calibration

alfiopuglisi:
Simulation can now be launched like this:

python main_scao.py 0 1 params_scao_plots.yml sncalib.yml 

with one or more additional yml files overriding/adding/removing objects from the main one


FabioRossiArcetri (approved):
OK

PR #34, opened 2024-10-11
Branch

alfiopuglisi:
Automatic trigger ordering based on input dependencies. Works also when multiple yaml param files are used.

Also uses the new :-1 syntax for delayed inputs, for example:

```
    layer_list: ['atmo.layer_list',
                  'pupilstop',
                  'dm.out_layer:-1']
```


FabioRossiArcetri (approved):
ok

PR #35, opened 2024-10-13
Streams

FabioRossiArcetri:


pyssata/processing_objects/pyr_slopec.py
```diff
@@ -72,39 +93,83 @@ def pupdata(self, p):
...
+        A = self.flat_pixels[self.pup_idx0]
+        B = self.flat_pixels[self.pup_idx1]
+        C = self.flat_pixels[self.pup_idx2]
+        D = self.flat_pixels[self.pup_idx3]
+        # Compute total intensity
+        # self.flux = self.xp.sum(A+B+C+D)
+        per_subap_sum = A+B+C+D
+        self.total_intensity = self.xp.sum(per_subap_sum)
+        clamp_generic_less(0, 0, self.total_intensity, xp=self.xp)
+        if self._norm_factor is not None:
```
alfiopuglisi: Here if self._norm_factor is not None, slopes will never be calculated.
Actually it was already wrong in the original code in pyr_compute_slopes.py, so maybe we can postpone the bugfix

pyssata/base_processing_obj.py
```diff
@@ -68,11 +68,23 @@ def prepare_trigger(self, t):
     def trigger_code(self):
         pass
 
+    def post_trigger(self):
```
alfiopuglisi, line 71: Should we put here a stream synchronize by default (if the graph was launched before) ?

PR #40, opened 2024-10-28
Optimize2

alfiopuglisi:
Several optimizations, 8m now 460Hz, ELT case 47Hz.

- using streams for function generators
- faster pyramid code
- using numpy for atmo evolution/propagation data except for the actual layers


FabioRossiArcetri (approved):
Looks good!

PR #42, opened 2024-10-29
atmo_random_phase class added

guidoagapito:


alfiopuglisi:
It would be better to also add a yml file that uses this new class, so that it can be tested

guidoagapito:
Done, see main/scao/params_atmo_random_phase.yml

PR #63, opened 2024-12-13
Polc

FabioRossiArcetri:


specula/processing_objects/modalrec.py
```diff
@@ -67,97 +67,30 @@ def __init__(self,
...
-
-    def trigger_code(self): # , slope_ptr=None
-        if self._recmat.recmat is None:
+        
+        
+        if self.polc:
+            self.out_comm = BaseValue('output commands from modal reconstructor', target_device_idx=target_device_idx)
+            self.inputs['in_commands'] = InputValue(type=BaseValue, optional=True)
+            self.inputs['in_commands_list'] = InputList(type=BaseValue, optional=True)
+            self.outputs['out_comm'] = self.out_comm
```
alfiopuglisi: This output "out_comm" is allocated but not set anywhere else. Should we remove it? Or is it intended for something that is not calculated at the moment?

FabioRossiArcetri: might be useful for debug purposes, but for now I would remove it.

alfiopuglisi (approved):

PR #73, opened 2025-01-13
Infinite atmo

FabioRossiArcetri:


specula/processing_objects/atmo_random_phase.py
```diff
@@ -75,31 +75,21 @@ def __init__(self,
...
-    def seed(self, value):
-        self._seed = value
-        self.compute()
-
-
-    def compute(self):
 
+    def initScreens(self):
         # Seed
+        self.seed = seed
```
alfiopuglisi: I think that this is a bug, "seed" is not defined in this context

specula/processing_objects/sh.py
```diff
@@ -19,7 +19,7 @@
 def abs2(u_fp, out, xp):
      out[:] = xp.real(u_fp * xp.conj(u_fp))
  
-rad2arcsec = 180 / np.pi * 3600
+RAD2ASEC = 180 / np.pi * 3600
```
alfiopuglisi: This line can be removed, RAD2ASEC is imported above

specula/processing_objects/atmo_random_phase.py
alfiopuglisi: I think that this is a bug, "seed" is not defined in this context

alfiopuglisi (commented):
In addition to the two comments in the code, checks are failing because the last version of main has not been merged to this branch. I suggest to to the merge (it will conflict in base_processing_obj.py because get_fft_plan() has been modified)

FabioRossiArcetri:
Now it should be ok

PR #82, opened 2025-02-27
(GA) low pass filter with example

guidoagapito:


alfiopuglisi:
I have put back BaseOperation to what it was in the main branch and it looks like it's working.
Maybe the series of changes in the end had no effect? In this case we can revert them

guidoagapito:
> I have put back BaseOperation to what it was in the main branch and it looks like it's working. Maybe the series of changes in the end had no effect? In this case we can revert them

The BaseOperation worked, but it overwrote the first input, which was visible in the plot.

guidoagapito:
Should the trigger_code method of plot_display.py become a trigger method?

alfiopuglisi:
I refactored the BaseOperation code, also using just "trigger" instead of "trigger_code", and added a test. It looks fine to me, let's wait for Fabio

FabioRossiArcetri (approved):

PR #91, opened 2025-03-11
(FRO) Splitting atmo propagation input layer list in two lists: the c…

FabioRossiArcetri:
…ommon layers (i.e. pupilstop, DMs) and the atmospheric layers (atmo layers different altitudes).

### DIFF
--- a/specula/processing_objects/atmo_propagation.py
+++ b/specula/processing_objects/atmo_propagation.py
@@ -54,7 +54,8 @@ class AtmoPropagation(BaseProcessingObj):
             ef.S0 = source.phot_density()
             self.outputs['out_'+name+'_ef'] = ef            
             
-        self.inputs['layer_list'] = InputList(type=Layer)
+        self.inputs['atmo_layer_list'] = InputList(type=Layer)
+        self.inputs['common_layer_list'] = InputList(type=Layer)
 
     def doFresnel_setup(self):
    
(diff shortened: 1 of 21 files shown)


specula/processing_objects/atmo_propagation.py
```diff
@@ -91,31 +92,26 @@ def field_propagator(*args, **kwargs):
...
-
-        layer_list = self.local_inputs['layer_list']
         for source_name, source in self.source_dict.items():
 
             output_ef = self.outputs['out_'+source_name+'_ef']
             output_ef.reset()
 
-            for layer in layer_list:
+            for layer in self.local_inputs['atmo_layer_list']:
+                interpolator = self.interpolators[source][layer]
```
alfiopuglisi, line 101: This code will not work if the interpolator is None, which can happen if an atmospheric layer is at height zero or a source is at infinity

specula/processing_objects/atmo_propagation.py
```diff
@@ -91,31 +92,26 @@ def field_propagator(*args, **kwargs):
...
-                        tempP = layer.phaseInNm
-                        tempP[tempA == 0] = self.xp.mean(tempP[tempA != 0])
-                        layer.phaseInNm = tempP
-
-                    output_ef.A *= interpolator.interpolate(layer.A)
-                    output_ef.phaseInNm += interpolator.interpolate(layer.phaseInNm)
+                output_ef.A *= interpolator.interpolate(layer.A)
+                output_ef.phaseInNm += interpolator.interpolate(layer.phaseInNm)
 
+            for layer in self.local_inputs['common_layer_list']:
```
alfiopuglisi, line 111: Same here. An interpolator will be needed if the layer is not at zero height, for example a DM conjugated at a certain height. Both the atmo list and the common layer list need the interpolator check

PR #96, opened 2025-03-26
(FRO) WIP versions of propagation and model analysis with per-layer r…

FabioRossiArcetri:
…esults to perform multi-layer calibrations.


FabioRossiArcetri:
@alfiopuglisi : are these contents already merged?

alfiopuglisi:
No they are not merged yet. However it needs another update from the main branch, since I see that there are conflicts

FabioRossiArcetri (commented):
@alfiopuglisi : now you can have a look at the way I resolved the conflicts

alfiopuglisi:
The conflict resolution is OK, but I see a problem: the new outputs are created in setup(), which is too late: they should be allocated in the constructor, otherwise they are not connected anywhere. If I try to use the new syntax in a YML file, I get an error on the outputs: ValueError: Object prop does does not have an output called out_on_axis_source_ef

FabioRossiArcetri:
@alfiopuglisi :conflicts are now solved and tests green, we can merge if you agree

alfiopuglisi (approved):

PR #97, opened 2025-03-27
(GA) Modal influence functions

guidoagapito:


FabioRossiArcetri (approved):

specula/lib/compute_zonal_ifunc.py
```diff
@@ -1,10 +1,27 @@
...
-                        obsratio=0.0, diaratio=1.0, mask=None):
+                        do_slaving=False, slaving_thr=0.1,
+                        obsratio=0.0, diaratio=1.0, mask=None, return_coordinates=False):
     """ Computes the ifs_cube matrix with Influence Functions using Thin Plate Splines """
+    
+    # Flag to check if we're using CuPy or another non-NumPy array library
+    is_cupy = xp.__name__ == "cupy"
+    
+    # Function to convert to NumPy array if needed
+    def to_numpy(arr):
```
alfiopuglisi, line 15: We already have these functions: cpuArray (from specula import cpuArray) is the equivalent of to_numpy(), and for to_xp() is it enough to use self.xp.array(arr) and it will be converted to GPU if it is used.

guidoagapito: https://github.com/FabioRossiArcetri/SPECULA/tree/zonalIfuncCpuArray

PR #103, opened 2025-04-02
(GA) _idx_finite removed and trigger reformatted

guidoagapito:


FabioRossiArcetri (commented):
I agree to simplify the code if the _idx_finite is not needed anymore, however before removing the numba optimization we should check the performances for this class it was shown that it runs faster on cpu+numba than on gpu.

guidoagapito:
I add another update, moving if sentences in prepare_ and post_trigger in the view of a better optimization with cuda streams

alfiopuglisi:
We have numba in the dependencies, and also some decorators (commented out and not really used anymore) in the specula init.py. I think we should remove everything, @FabioRossiArcetri what do you think?

PR #105, opened 2025-04-10
(GA) bugfix for pupdata subdir

guidoagapito:


specula/calib_manager.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, root_dir):
             'sn': 'slopenulls/',
             'background': 'backgrounds/',
             'pupils': 'pupils/',
-            'pupdata': 'pupils',
+            'pupdata': 'pupils/',
```
alfiopuglisi, line 26: Are the slashes really needed? The code in self.joinpath() will do an os.path.join so it should not matter if there is a slash or not. Which routine is failing without the slash?

guidoagapito: Ok, the issue I have comes from something else, so I close this pull request.

PR #108, opened 2025-04-12
(GA) test for calibration and simulation of a SH SCAO toy system

guidoagapito:
This pull requests includes changes to:

- calib_manager to include subdir for slopenulls
- slopes to manage the save of data with pupdata_tag or subapdata_tag
- atmo_propagation where the layers_list are optional
- simul to handle optional and union types in build_objects


specula/processing_objects/atmo_propagation.py
FabioRossiArcetri, line 1: what if both optional inputs are not provided? I guess it should be checked and reported as an error

guidoagapito: You are right and I think that common_layer_list should not be set as optional, because at least a pupilstop is required.

FabioRossiArcetri (approved):

alfiopuglisi (approved):

PR #113, opened 2025-04-18
Multi-wavelength PSF and nested TN

alfiopuglisi:
Implemented nested TN: saving additional data from a replay TN, creating a new TN inside it, using the --nested-TN option in main_simul.py

Added wavelengthInNm_list option to PSF object, allowing multiple PSFs to be computed in a single object.


FabioRossiArcetri (commented):
I guesse it should be 'store_false', or am I wrong?

alfiopuglisi:
`I guesse it should be 'store_false', or am I wrong?`

'store_true' only stores True if the command line option is given, otherwise is has a value of False, I think it is correct as it is.

PR #115, opened 2025-04-22
(GA) type for ifunc in init and smart inversion of ifunc

guidoagapito:


alfiopuglisi:
If we need to support both normal and inverse influence functions, then maybe it is better to modify the IFunc data object to be able to hold both, for example just adding a flag that specifies whether the data is the influence function or its inversion, and checking the flag inside the inverse() method.

guidoagapito:
How can we implement it?
In the parameters file we have:

```
modal_analysis:
  class:            'ModalAnalysis'
  ifunc_object:     'MAORY_480pix_41nacts_cir_spider2023_5zern_10000.0cn_inv'
  dorms:            True
  inputs:
    in_ef: 'atmo.out_on_axis_source_ef'
  outputs: ['out_modes']
```
  
The ifunc object is built by restoring the data from disk and no additional parameter is passed to it before initialising the modal_nalaysis object.
I do not see how a flag can be passed from outside, but perhaps it can be automated by comparing the shape of self._influence_function and self._idx_inf_func.

alfiopuglisi:
If the standard is "tag.fits"and "tag_inv.fits", we could give the tag for the influence function and the IFunc method could load one or the other (or both) depending if one asks for the direct or inverse. Given the sizes of these matrices, maybe it would be better to implement it using lazy loading, in order to avoid loading one of the two that is not going to be used.

guidoagapito:
So you suggest that the ifunc object is initialised with without reading the file from disk, and only when passed to the processing object is the array loaded from disk?
In this way it can load the required one, or if inverse is not present it will read the direct one and invert it on the fly (and store the inverse on disk).

PR #121, opened 2025-05-05
Intmat saving bugfix and test

alfiopuglisi:


guidoagapito:
I think that self._intmat should be self.intmat or a method should be added to access the data in the object.

alfiopuglisi:
Ok changed to self.intmat

guidoagapito (approved):

PR #125, opened 2025-05-08
Common params ref

FabioRossiArcetri:


guidoagapito (approved):

specula/data_objects/pupilstop.py
```diff
@@ -20,7 +19,11 @@ def __init__(self,
                  target_device_idx: int=None,
                  precision: int=None):
 
-        super().__init__(pixel_pupil, pixel_pupil, pixel_pitch, height=0,
+        self.simul_params = simul_params
+        self.pixel_pupil = self.simul_params.pixel_pupil
+        self.pixel_pitch = self.simul_params.pixel_pitch
+
+        super().__init__(self.pixel_pupil, self.pixel_pupil, self.pixel_pitch, height=0,
```
alfiopuglisi, line 26: Do we need to use "self.pixel_pupil", etc? Since it is only used a few times, there is no need to create all these variables, maybe it's simpler to just use "simul_params.pixel_pupil"

specula/data_objects/simul_params.py
```diff
@@ -0,0 +1,49 @@
...
+    time_step : float
+        The duration of a single timestep in seconds (number of timesteps = int(total_time/time_step) )
+    zenithAngleInDeg : float
+        The zenith angle of the telescope
+    display_server : bool
+
+    '''
+
+    def __init__(self,
+                 pixel_pupil: int = 120,
```
alfiopuglisi, line 26: I would remove the default values for pixel_pupil and pixel_pitch, since they really need to be specified. Otherwise we default to a 6m telescope :) While the other values are reasonable

specula/processing_objects/atmo_propagation.py
```diff
@@ -166,15 +174,15 @@ def layer_interpolator(self, source, layer):
         half_pixel_layer -= layer.shiftXYinPixel
 
         if self.pupil_position is not None and pixel_layer > self.pixel_pupil_size and np.isinf(source.height):
-            pixel_position_s = source.r * layer.height / layer.pixel_pitch
+            pixel_position_s = source.r * layer.height * self.airmass / layer.pixel_pitch
```
alfiopuglisi, line 177: layer.height was not scaled by airmass before. Was it a bug that is now fixed?

specula/processing_objects/iir_filter.py
```diff
@@ -21,6 +24,10 @@ def __init__(self, iir_filter_data: IirFilterData,
                  precision=None
                  ):  
 
+        self.simul_params = simul_params
+        self.time_step = self.simul_params.time_step
+        print('self.simul_params', self.simul_params)
```
alfiopuglisi: These print(self.simul_params) can be removed, there is more than one

specula/processing_objects/low_pass_filter.py
```diff
@@ -16,7 +16,11 @@ def __init__(self,
                  precision: int=None
                 ):
         
-        samp_freq = 1 / time_step
+        self.simul_params = simul_params
+        self.time_step = self.simul_params.time_step
+        print('self.simul_params', self.simul_params)
+        
+        samp_freq = 1 / self.time_step
```
alfiopuglisi, line 23: Same comment as before, this could be just "samp_freq = 1 / simul_params.time_step" and the previous lines removed.

specula/simul.py
```diff
@@ -134,14 +136,22 @@ def trigger_order(self, params_orig):
         return order, order_index
 
 
+    def setSimulParams(self, params):
+        for key, pars in params.items():
+            #if 'class' in pars:
```
alfiopuglisi: Remove the commented out "if"

specula/simul.py
```diff
@@ -233,6 +246,8 @@ def build_objects(self, params):
                 my_params['info_getter'] = self.get_info
 
             my_params.update(pars2)
+            print(key)
+            print(my_params)
```
alfiopuglisi: Other debug prints to be removed

specula/simul.py
```diff
@@ -216,7 +226,10 @@ def build_objects(self, params):
                     pars2[name] = value
 
             # Add global and class-specific params if needed
-            my_params = {k: main[k] for k in args if k in main}
+            # my_params = {k: main[k] for k in args if k in main}
```
alfiopuglisi: Let's remove the this unused code, even if commented out, and just leave the new version

specula/data_objects/simul_params.py
alfiopuglisi: If we want to keep a default value in case an object does not use pixel_pupil and pixel_pitch at all, then let's set None as the default so that it gives an error when used without having been set.

specula/processing_objects/atmo_infinite_evolution.py
```diff
@@ -245,6 +242,11 @@ def __init__(self,
 
         super().__init__(target_device_idx=target_device_idx, precision=precision)
         
+        self.simul_params = simul_params
+        print('self.simul_params', self.simul_params)
```
alfiopuglisi: Remove print

specula/processing_objects/atmo_propagation.py
```diff
@@ -27,18 +27,22 @@ def __init__(self,
 
         super().__init__(target_device_idx=target_device_idx, precision=precision)
 
+        self.simul_params = simul_params
+        print('self.simul_params', self.simul_params)
```
alfiopuglisi: Remove print

specula/processing_objects/dm.py
```diff
@@ -29,6 +30,10 @@ def __init__(self,
                  ):
         super().__init__(target_device_idx=target_device_idx, precision=precision)
 
+        self.simul_params = simul_params
+        self.pixel_pitch = self.simul_params.pixel_pitch
+        print('self.simul_params', self.simul_params)
```
alfiopuglisi: Remove print

specula/processing_objects/atmo_propagation.py
FabioRossiArcetri: The scaling with the airmass was done in the Source class code, which did not make much sense.

specula/processing_objects/iir_filter.py
FabioRossiArcetri: agreed removing all these prints.

specula/simul.py
FabioRossiArcetri: yes

specula/simul.py
FabioRossiArcetri: yes

specula/data_objects/pupilstop.py
FabioRossiArcetri: For the cost, I would prefer to have a standard behavior where all the needed simul_params members are copied this way, independently from the fact they are going to be used in the __init__ method or in other methods.

alfiopuglisi (approved):

PR #126, opened 2025-05-09
(GA) kernels: most of the parameters are now arguments of the init method

guidoagapito:


specula/data_objects/gaussian_convolution_kernel.py
```diff
@@ -63,17 +94,18 @@ def restore(filename, target_device_idx=None, kernel_obj=None, return_fft=False)
...
+            kernel_obj.pxscale = hdr['PXSCALE']
+            kernel_obj.dimension = hdr['DIM']
+            kernel_obj.oversampling = hdr['OVERSAMP']
+            kernel_obj.positive_shift_tt = hdr['POSTT']
 
         # Read the kernel data from extension 1
-        kernel_obj.real_kernels = kernel_obj.xp.array(fits.getdata(filename, ext=1))       
+        data = kernel_obj.xp.array(fits.getdata(filename, ext=1))
+        if kernel_obj.real_kernels.shape == data.shape and kernel_obj.real_kernels.dtype == data.dtype:
+            kernel_obj.real_kernels[...] = data
```
alfiopuglisi, line 107: Even if this works, I think that the current syntax is "real_kernels[:] = data" using a colon rather than the three dots. Same think in the other class

guidoagapito: Both syntaxes are correct, but have slightly different meanings:

kernel_obj.real_kernels[...] is called ‘ellipsis indexing’ and assigns date values to the entire existing array, regardless of the number of dimensions. It works well with multi-dimensional arrays.

kernel_obj.real_kernels[:] is a slicing that assigns date values to the existing array, but is most commonly used for one-dimensional arrays.

PR #131, opened 2025-05-12
Simpler setup() method

alfiopuglisi:
Simplified setup() method in processing objects, now that SimulParams is available.
Also fixed a GPU bug in FuncGenerator related to time handling in GPU.


guidoagapito (approved):

specula/base_processing_obj.py
FabioRossiArcetri, line 1: Does this modification implies that we must pass the SimulParams to all the processing objects __init__ method?

FabioRossiArcetri (approved):

PR #132, opened 2025-05-12
(GA) Handle int_gain and ff setup based on n_mode argument

guidoagapito:


specula/processing_objects/integrator.py
```diff
@@ -8,14 +8,40 @@ class Integrator(IirFilter):
...
+        Integrator class for processing signals using an IIR filter.
+        This class is a specialized version of the IirFilter class, designed to handle
+        integration operations with specific gain and forgetting factor settings.
+        """
+        # Handle gain (int_gain) and forgetting factor (ff) setup based on n_modes:
+        # - If n_modes is provided, it specifies how many modes (channels) to use.
+        # - If n_modes is an integer, convert it to a list for uniform processing.
+        # - If n_modes is a list, its length must match int_gain.
+        # - Each int_gain[i] is expanded into a block of size n_modes[i].
+        #   Example: n_modes=[2,3], int_gain=[0.5, 1.0] -> int_gain = [0.5, 0.5, 1.0, 1.0, 1.0]
```
alfiopuglisi, line 29: We can also say that this behaviour (repeat each element of ff by the corresponding number in n_modes) is the same as np.repeat:
```
>>> np.repeat([0.5, 1.0], [2,3])
array([0.5, 0.5, 1. , 1. , 1. ])
```

guidoagapito: I added a comment.

PR #134, opened 2025-05-12
(GA) SH noise covariance matrix considering elongated / truncated spots

guidoagapito:


specula/lib/calc_noise_cov_elong.py
```diff
@@ -0,0 +1,304 @@
...
+
+def calc_noise_cov_elong(diameter_in_m, zenith_angle_in_deg, na_thickness_in_m, launcher_coord_in_m,
+                         sub_aps_index, n_sub_aps, sub_aps_fov, sh_spot_fwhm, sigma_noise2,
+                         t_g_parameter, h_in_m=None, user_pofile_xy=None, theta=None,
+                         only_diag=False, eta_is_not_one=False, display=False, verbose=False):
+    """
+    Computes noise covariance matrix considering WFS sub-aperture, laser launcher and sodium layer geometry.
+    
+    Parameters:
+        diameter_in_m (float): Telescope diameter in meters
```
alfiopuglisi, line 17: If this routine does not use cupy (there is no xp argument, and it uses np explicitly inside), then I suggest to convert all float/array/list arguments using cpuArray(), because if called from specula objects, it could get 
 some cupy arrays. So for example:
`diameter_in_m = cpuArray(diameter_in_m)`
etc

guidoagapito: Done

alfiopuglisi (approved):

PR #137, opened 2025-05-14
(FRO) Printing GPU memory usage for each processing object before the…

FabioRossiArcetri:
… simulation starts.


specula/base_processing_obj.py
```diff
@@ -160,7 +160,7 @@ def verbose(self):
     @verbose.setter
     def verbose(self, value):
         self._verbose = value
-
+
```
guidoagapito: This change should be discarded.

alfiopuglisi (approved):

PR #140, opened 2025-05-15
(GA) Code from sh prepare_kernels moved to kernel classes

guidoagapito:


specula/data_objects/gaussian_convolution_kernel.py
```diff
@@ -2,6 +2,7 @@
 from specula.data_objects.convolution_kernel import ConvolutionKernel, lgs_map_sh
 from specula import cpuArray
 
+import os
```
alfiopuglisi: Not needed?

guidoagapito: Yes, I removed the not used imports.

PR #141, opened 2025-05-15
(GA) Vibration class

guidoagapito:


specula/processing_objects/func_generator.py
```diff
@@ -4,37 +4,93 @@
...
 class FuncGenerator(BaseProcessingObj):
     def __init__(self,
-                 func_type='SIN', 
-                 nmodes: int=None, 
-                 time_hist=None, 
-                 psd=None, 
-                 fr_psd=None, 
-                 continuous_psd=None, 
-                 constant: list=None, 
+                 simul_params: SimulParams=None,
```
alfiopuglisi, line 161: Up to now the simul_params argument is always mandatory, while here it will be required only in case the FuncGenerator is of type VIB_PSD. @FabioRossiArcetri  what do you think?
An alternative way would be to have a VibrationGenerator separated from FuncGenerator, not even derived, since the parameter list and the generation code is almost completely different.

guidoagapito: I set this as mandatory to avoid adding it in many cases where it is not required.
We could have a vibration class derived from func_generator.

FabioRossiArcetri: I would also prefer an implementation of the vibration generator as a separate object.

FabioRossiArcetri (approved):

FabioRossiArcetri (approved):

PR #142, opened 2025-05-15
(GA) modalrec with implicit POLC

guidoagapito:


specula/processing_objects/modalrec_implicit_polc.py
```diff
@@ -0,0 +1,125 @@
...
+        del comm_mat
+        
+        # set up the H matrix
+        h_mat = self.comm_mat.recmat @ self.intmat.intmat
+        h_mat = self.xp.identity(h_mat.shape[0], dtype=self.dtype) - h_mat
+        self.h_mat = Recmat(h_mat, target_device_idx=target_device_idx, precision=precision)
+        # Now self.intmat can be removed to save memory
+        self.intmat = None
+        # delete the H matrix to save memory
+        del h_mat
```
alfiopuglisi: comm_mat and h_mat have an additional reference inside the Recmat objects, so there is no benefit in deleting them here - the total memory is the same.

guidoagapito: del removed.

alfiopuglisi (approved):

PR #145, opened 2025-05-15
Dm npixels

alfiopuglisi:


guidoagapito (commented):
This is not true in general, self.simul_params.pixel_pupil can be the default value, but we need a way to override it if needed.

alfiopuglisi:
> This is not true in general, self.simul_params.pixel_pupil can be the default value, but we need a way to override it if needed.

Ok, now it is basically as before, except that npixels is not given and cannot be derived from masks or ifunc, it is set to pixel_pupil. Also I have added sanity checks for parameters. @guidoagapito

guidoagapito (approved):

FabioRossiArcetri:
@alfiopuglisi can you resolove the conflicts? Looks like it's just the to_xp integration.

alfiopuglisi:
They are fixed now

FabioRossiArcetri (approved):

PR #146, opened 2025-05-16
(GA) compatibility with PASSATA in m2c

guidoagapito:


alfiopuglisi:
In params_morfeo_ltao.yml we have beeing using an m2c from PASSATA without problems:

```
dm:
  class:             'DM'
  simul_params_ref:  'main'
  ifunc_object:      'M4_eso_zonal_486p_5352a_slaved'
  m2c_object:        'M4_eso_39.5m_486p_20230215'
```

 If it had been transposed, I don't think that the loop would be able to close...

guidoagapito:
You are right.

PR #156, opened 2025-06-05
Morfeo full

alfiopuglisi:
Fixed MORFEO LGS rotation, tomographic matrices, number of modes.
Revised handling of mode ranges in influence functons, m2c and modalrec.
Added exception catches with additional debug output
BaseOperation and LinearCombination classes are now using pre-allocated GPU arrays
DataStore is now able to save Intensity data objects


guidoagapito (approved):

alfiopuglisi:
The mode range handling in m2c, dm and influence functions has become quite complex. We will need to add some documentation and tests...

FabioRossiArcetri (approved):

PR #177, opened 2025-06-27
Mpi support

FabioRossiArcetri:


specula/base_processing_obj.py
```diff
@@ -66,15 +74,22 @@ def prepare_trigger(self, t):
         if self.target_device_idx >= 0:
             self._target_device.use()
 
+        if self.target_device_idx >= 0:
+            self._target_device.use()
+
```
alfiopuglisi, line 79: This is a duplication of the previous 3 lines

specula/base_processing_obj.py
```diff
@@ -175,9 +204,12 @@ def setup(self):
         """
         if self.target_device_idx >= 0:
             self._target_device.use()
-        for name, input in self.inputs.items():
-            if input.get(self.target_device_idx) is None and not input.optional:
-                raise ValueError(f'Input {name} for object {self} has not been set')
+        for input_name, input in self.inputs.items():
```
alfiopuglisi, line 230: This loop could be replaced by a call to BaseProcessingObj.prepare_trigger(), which perform get() on all inputs and updates local_inputs, and also handles list types correctly. Or maybe we can factor out the loop (the version in prepare_trigger) and call it from both setup() and prepare_trigger(). The check on optional inputs must be kept somehow

specula/connections.py
```diff
@@ -1,80 +1,146 @@
...
+            print(process_rank, 'received obj type', type(output_data), flush=True)
+            print(process_rank, 'received obj', output_data, flush=True)
+
+            #if self.cloned_value is None:
+            # update copyTo to handle same target_device_idx but different rank
+            self.cloned_value = output_data.copyTo(target_device_idx)
+            print(process_rank, 'Received data copied', flush=True)
+            #else:
+                # update transferDataTo to handle same target_device_idx but different rank
+            #output_data.transferDataTo(self.cloned_value)
```
alfiopuglisi: If we want to use streams in processing objects, using transferDataTo() is mandatory otherwise streams will reference the wrong memory area

specula/loop_control.py
```diff
@@ -59,18 +65,74 @@ def start(self, run_time, dt, t0=0, stop_on_data=None, stop_at_time=None,
...
+
+
+        print(process_rank, 'at barrier A', flush=True)
+        process_comm.barrier()
+        print(process_rank, 'after barrier A', flush=True)
+
+        print(process_rank, 'self._max_order', self._max_order, flush=True)
+
+        '''
+        for i in range(self.max_global_order+1):
```
alfiopuglisi: What is the purpose of this loop? I thought that the previous loop with send_outputs() would be enough for calling setup() afterwards

specula/simul.py
```diff
@@ -61,27 +75,43 @@ def output_ref(self, output_name):
...
+                    return None, obj_name
+                else:
+                    raise ValueError(f'Object {obj_name} does not exist anywhere')
             if not attr_name in self.objs[obj_name].outputs:
                 raise ValueError(f'Object {obj_name} does not define an output with name {attr_name}')
             output_ref = self.objs[obj_name].outputs[attr_name]
         else:
-            output_ref = self.objs[output_name]
-        return output_ref
+            # this mus be a data object
```
alfiopuglisi: Are we allowing remote inputs from remote data objects? It could be too much.. maybe we can limit to processing objects.

specula/simul.py
```diff
@@ -26,10 +36,14 @@ def __init__(self,
...
                  diagram_title=None,
-                 diagram_filename=None                 
+                 diagram_filename=None
                  ):
         if len(param_files) < 1:
             raise ValueError('At least one Yaml parameter file must be present')
+        self.all_objs_ranks = {}
+        self.remote_objs_ranks = {}
+        self.remote_objs_types = {}
+        self.remote_objs_params = {}
```
alfiopuglisi: self.remote_objs_params is not used later.

specula/simul.py
```diff
@@ -278,28 +317,46 @@ def build_objects(self, params):
...
-            try:
-                self.objs[key] = klass(**my_params)
-            except Exception:
-                print(f'Exception building', key)
-                raise
-            if classname != 'SimulParams':
-                self.objs[key].stopMemUsageCount()
+            # create the simulations objects for this process. Data Object and SimulParams are always
+            # created, no matter what their rank (assigned process) is.
+            if process_rank==target_rank or issubclass(klass, BaseDataObj) or classname=='SimulParams':
```
alfiopuglisi: The check for process_rank==target_rank could be done much earlier: since remote_objs_params is not used, for remote objects we only need "target_rank" and "klass", so most of the initialization can be skipped.

main/scao/main_mpi.py
```diff
@@ -0,0 +1,80 @@
+# mpiexec -n 2 python script.py args
+
+from mpi4py import MPI
```
alfiopuglisi: We need to add mpi4py as a dependency in setup.py

FabioRossiArcetri: Yes, should we make it optional and organize the code accordingly?

alfiopuglisi: If everything is on pip, maybe it's better to have a simpler installation with everything and then have command line switches. But I don't know if it is possible to install everything automatically (on my account on gandalf it did not work).
Otherwise, lets' make it optional and organize the code.

specula/base_processing_obj.py
FabioRossiArcetri: Does the work done on this topic include the optional parameters handling?

specula/simul.py
FabioRossiArcetri: ok, removing it

specula/base_processing_obj.py
alfiopuglisi: The optional inputs are a problem at the moment, because we should avoid calling get() for remote inputs (it would block because no one has done a corresponding send)

specula/simul.py
FabioRossiArcetri, line 1: I guess this way I solved the conflict makes the tests fail, while for both the main and the mpiSupport branches alone the tests are ok, @alfiopuglisi do you have an idea of what might have happened here?

guidoagapito: The issues comes from `build_order` method of simul.py:

```
            pars = params[key]
            for name, value in pars.items():
                if name.endswith('_ref'):
                    objlist = value if type(value) is list else [value]
                    for output in objlist:
                        owner = self.output_owner(output)
                        if owner not in build_order:
                            add_to_build_order(owner)
```

output is `'main'` and `self.output_owner` fails.

`output` is `'main'` because we have sections of the yaml file with this line:

`simul_params_ref:     'main'`

PR #185, opened 2025-07-02
(GA) vectorized version of extrapolation

guidoagapito:


alfiopuglisi (approved):

alfiopuglisi (commented):
I spoke too soon, a few tests are failing on gandalf with pytest on GPU

guidoagapito:
Now it is working. I added the ".copy()" I removed in the first commit and I have also change the way the output is set so that we have the same implementation of the previous extrapolation.

PR #188, opened 2025-07-03
(GA) emccd_gain valueError becomes a warning

guidoagapito:


specula/processing_objects/ccd.py
```diff
@@ -103,7 +103,7 @@ def __init__(self,
                 self._ADU_gain = 1 / 20
         else:
             if emccd_gain is not None:
-                raise ValueError('emccd_gain must be None if excess_noise is False')
+                print('ATTENTION: emccd_gain will not be used if excess_noise is False')
```
alfiopuglisi: For this kind of messages, we could use the standard warnings module. This would become:
`warnings.warn('emccd_gain will not be used if excess_noise is False', warnings.RuntimeWarning)`

alfiopuglisi (approved):

PR #190, opened 2025-07-04
(GA) Plate scale coefficients for MCAO correction from DMs

guidoagapito:


specula/lib/platescale_coeff.py
```diff
@@ -0,0 +1,115 @@
...
+        Dictionary containing plate scale parameters
+    """
+
+    # plate scale modes are 3: focus and 2 astigmatism modes
+    n_modes_ps = 3
+
+    # Extract influence functions from the first DM
+    aIfunc = dm_list[0]._ifunc.influence_function
+    # The first DM is the ground DM, so we always skip the tip-tilt modes
+    if hasattr(dm_list[0], 'm2c') and dm_list[0].m2c is not None:
```
alfiopuglisi: a DM object always has the self.m2c attribute, there should be no need to check with hasattr()

specula/lib/platescale_coeff.py
```diff
@@ -0,0 +1,115 @@
...
+    -------
+    plateScale : dict
+        Dictionary containing plate scale parameters
+    """
+
+    # plate scale modes are 3: focus and 2 astigmatism modes
+    n_modes_ps = 3
+
+    # Extract influence functions from the first DM
+    aIfunc = dm_list[0]._ifunc.influence_function
```
alfiopuglisi: a DM object has a "ifunc" property that can be used in place of "._ifunc.influence_function"

alfiopuglisi (approved):

PR #193, opened 2025-07-04
(GA) first extension is phase not mask

guidoagapito:


guidoagapito:
This will break compatibility with PASSATA files, but it is necessary for consistency.

alfiopuglisi:
Consistency with what? Even for EF objects, the first extension is the amplitude mask, and the second one is the phase

guidoagapito:
Pupilstop save method calls super save that is layers that in turn calls electric_field.
The file saved by it looks like this:

```
In [6]: hdul.info()
Filename: /Users/guido/GitHub/SPECULA_scripts/calibration/pupilstop/mask_160p_4petals.fits
No.    Name      Ver    Type      Cards   Dimensions   Format
  0  PRIMARY       1 PrimaryHDU      13   (160, 160)   uint8   
  1                1 ImageHDU         7   (160, 160)   float64   
  2                1 ImageHDU         7   (160, 160)   uint8   
  3                1 ImageHDU         6   (2,)   int64   
  4                1 ImageHDU         6   (1,)   float64  
``` 
  
There is two copies of the mask, No. 0 and 2.
  
This is why:
  
```
        input_mask = fits.getdata(filename, ext=1)
        dim = fits.getdata(filename, ext=2)
        pixel_pitch = fits.getdata(filename, ext=3)[0]
```
        
is wrong, and

```
         input_mask = fits.getdata(filename, ext=2)
        dim = fits.getdata(filename, ext=3)
        pixel_pitch = fits.getdata(filename, ext=4)[0]
```
        
works.
Probably we should change the save method and update update restore method.
Do you have some suggestions?

alfiopuglisi:
The save/restore methods must be revised everywhere, let's discuss this
together we we are back

FabioRossiArcetri (approved):

PR #196, opened 2025-07-07
Allow free build order in the YAML files

alfiopuglisi:
Currently, objects references by other ones (like sources and DM influence functions) must be declared in the YAML files before they are used.
This PR allows any definition order in the YAML file.


guidoagapito (approved):

guidoagapito:
Three tests files because:

"FAILED test/test_field_analyser.py::TestShSimulation::test_field_analyser_psf - PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'D:\\a\\SPECULA\\SPECULA\\test\\data\\20250710_165226\\comm.fits'"

This happens when the tearDown method of TestShSimulation is called.

I think this error could be solved with the fix proposed in https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/200

alfiopuglisi:
Yes it is fixed by #200, the merge needs to be done in order: first #200 then this one

PR #199, opened 2025-07-10
New specula entry point Specula

alfiopuglisi:
- Added a "specula" command, replacing main_simul.py and main_mpi.py, installed as an entry point so it is included in the path (no need to be in main/scao).
- main_simul is available from python code as specula.main_simul to easily run simulations from python (see tutorial)


main/scao/runAll.sh
guidoagapito, line 1: typo:

pspecula arams_atmo_random_phase.yml

should be

specula params_atmo_random_phase.yml

guidoagapito:
docs/tutorials/step_response_tutorial.rst should be updated

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #200, opened 2025-07-10
Fixed bug in DataSource

alfiopuglisi:
FITS files were not closed correctly and tests on Windows sometimes fail

### DIFF
--- a/.github/workflows/pythonpackage.yml
+++ b/.github/workflows/pythonpackage.yml
@@ -31,8 +31,8 @@ jobs:
     - name: Install dependencies
       run: |
         bash
-        sudo apt update
         if ${{runner.os == 'Linux'}}; then
+          sudo apt update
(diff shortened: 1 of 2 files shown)


guidoagapito (approved):

guidoagapito:
Tests on Ubuntu are failing because "Installing dependencies" part gets the error "Unable to fetch some archives".
The failure occurred because the workflow tried to install dependencies using apt, but some required packages could not be fetched (404 Not Found), specifically:

uuid-dev
libblkid-dev
libmount-dev

Copilot suggest to run apt-get update before installing packages.

alfiopuglisi:
Tests are fixed now after adding "sudo apt update", strange that it only came up in this branch and nowhere else

FabioRossiArcetri (approved):

PR #207, opened 2025-07-22
(FRO) Add n_simul parameter to specula to perform n simulations in se…

FabioRossiArcetri:
…quence. The remove and override directives in the yaml files have now also the _simIdx syntax to specify changes that should be done only for a specific simulation of the sequence.

### DIFF
--- a/main/scao/calib_morfeo_small_rec.yml
+++ b/main/scao/calib_morfeo_small_rec.yml
@@ -1,0 +1,75 @@
+
+
+pushpull:
+  class:     'FuncGenerator'
+  func_type: 'PUSHPULL'
+  nmodes:     100
+  vect_amplitude:  [50, 50, 50, 50, 50, 50, 50, 50, 50, 50, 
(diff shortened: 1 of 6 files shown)


alfiopuglisi:
Since the parameters contain multiple numbers, like: `im_calibrator1_override_2` or `dm2_override_2`, maybe we could have a more explicit syntax? Like for example `dm2_override_simul2` so that it is clear that the "2" refers to the simulation index

FabioRossiArcetri:
> Since the parameters contain multiple numbers, like: `im_calibrator1_override_2` or `dm2_override_2`, maybe we could have a more explicit syntax? Like for example `dm2_override_simul2` so that it is clear that the "2" refers to the simulation index

We can easily do this, but consider that the _number is an extension of "remove" or "override", so it should be already apparent that the number in these cases refers to the simulation index. I would agree of we also plan to add creation of objects conditional to the simulation index.

alfiopuglisi (approved):

alfiopuglisi:
Ok we can leave it as it is now

PR #208, opened 2025-07-22
(GA) Pyramid optical gain estimator and IIR filter gain optimizer

guidoagapito:


specula/processing_objects/demodulator.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def prepare_trigger(self, t):
+        super().prepare_trigger(t)
+        self.input = self.local_inputs['in_data']
+
+    def trigger_code(self):
+        t = self.current_time
+
+        # Store data if input is ready
+        if self.input.generation_time == t:
```
alfiopuglisi: This "`if self.input.generation_time == t`" should not be necessary, trigger_code is called only when the input is ready

guidoagapito: I removed the if statement

specula/processing_objects/demodulator.py
```diff
@@ -0,0 +1,214 @@
...
+            # Extract data for the specified modes
+            if self.input.value.ndim > 1:
+                # Multi-dimensional data - extract modes
+                mode_data = self.input.value[self.mode_numbers]
+            else:
+                # 1D data
+                mode_data = self.input.value
```
alfiopuglisi: If this is modified to:
`mode_data = [self.input.value]`
then data_array later on will always be 2d and the if (ndim==2) can be removed.

specula/processing_objects/demodulator.py
```diff
@@ -0,0 +1,212 @@
...
+    def trigger_code(self):
+        t = self.current_time
+
+        # Extract data for the specified modes
+        if self.input.value.ndim > 1:
+            # Multi-dimensional data - extract modes
+            mode_data = self.input.value[self.mode_numbers]
+        else:
+            # 1D data
+            mode_data = self.input.value
```
alfiopuglisi, line 59: If this is modified to:
`mode_data = [self.input.value]
`then data_array later on will always be 2d and the "else (Single mode") part of the demodulation can be removed.

specula/processing_objects/optical_gain_estimator.py
```diff
@@ -0,0 +1,227 @@
...
+
+        self.output.value = output
+        self.output.generation_time = t
+
+        if self.verbose:
+            print(f'Optical gain output: {output}')
+
+    def _evaluate_expressions(self):
+        """
+        Evaluate expressions for advanced output mapping.
```
alfiopuglisi: If we want really generic expressions we could use eval(), but it might give too much freedom

specula/processing_objects/demodulator.py
guidoagapito: Done, other changes were required.

specula/processing_objects/optical_gain_estimator.py
guidoagapito: I removed this part because it was never used in PASSATA and I don't have any test.

alfiopuglisi (approved):

FabioRossiArcetri (approved):

Issue #232, opened 2025-07-30
bug with ef_at_lambda when out size is bigger than phi


guidoagapito:
When out size is bigger than phi I get this error:

```
Traceback (most recent call last):
  File "/home/guido/.conda/envs/tiptop/bin/specula", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/scripts/specula.py", line 23, in main
    specula.main_simul(**vars(args))
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/__init__.py", line 261, in main_simul
    ).run()
      ^^^^^
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/simul.py", line 736, in run
    self.loop.run(run_time=self.mainParams['total_time'], dt=self.mainParams['time_step'], speed_report=True)
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/loop_control.py", line 51, in run
    self.iter()
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/loop_control.py", line 134, in iter
    element.check_ready(self._t)
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/base_processing_obj.py", line 192, in check_ready
    self.prepare_trigger(t)
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/processing_objects/modulated_pyramid.py", line 442, in prepare_trigger
    self._wf_interpolated.ef_at_lambda(self.wavelength_in_nm, out=self.ef)
  File "/raid1/guido/pythonLib/guidoSPECULA/specula/data_objects/electric_field.py", line 85, in ef_at_lambda
    ef = self.xp.exp(1j * phi, dtype=self.complex_dtype, out=out)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cupy/_core/_kernel.pyx", line 1314, in cupy._core._kernel.ufunc.__call__
  File "cupy/_core/internal.pyx", line 381, in cupy._core.internal._broadcast_core
ValueError: operands could not be broadcast together with shapes (400, 400) (800, 800)
```

The method is

```
    def ef_at_lambda(self, wavelengthInNm, slicey=None, slicex=None, out=None):
        if slicey is None:
            slicey = np.s_[:]
        if slicex is None:
            slicex = np.s_[:]
[...]

PR #233, opened 2025-07-31
Rotation and interpolation compatibility

alfiopuglisi:
Should fix issue #232


specula/processing_objects/modulated_pyramid.py
```diff
@@ -504,13 +504,18 @@ def setup(self):
         in_ef = self.local_inputs['in_ef']
 
         # Determine if interpolation is needed (like in SH)
-        if self.rotAnglePhInDeg != 0 or self.xShiftPhInPixel != 0 or self.yShiftPhInPixel != 0:
+        if (self.fft_sampling != in_ef.size[0] or
```
guidoagapito: I think we can use self.fov_res

guidoagapito:
I think the tests failed for the change of sign in rotation.

specula/processing_objects/modulated_pyramid.py
```diff
@@ -519,8 +524,8 @@ def setup(self):
             # Create the interpolator (like in SH)
             self.interp = Interp2D(
                 in_ef.size,
-                in_ef.size,
-                self.rotAnglePhInDeg,
+                (self.fft_sampling, self.fft_sampling),
+                -self.rotAnglePhInDeg,  # Negative angle for PASSATA compatibility
```
guidoagapito, line 527: I think this is not coherent with SH

guidoagapito (approved):

specula/processing_objects/modulated_pyramid.py
alfiopuglisi: Ok done

specula/processing_objects/modulated_pyramid.py
alfiopuglisi: SH angles have also been inverted in #237

Issue #80, opened 2025-02-26
bug with L0 as vector


guidoagapito:
When L0 in AtmoEvolution is a vector I get this error:

```
(tiptop) guido@finrod scao % python main_simul.py params_scao_plots.yml
...
Traceback (most recent call last):
  File "/Users/guido/GitHub/SPECULA/main/scao/main_simul.py", line 23, in <module>
    simul.run()
  File "/Users/guido/GitHub/SPECULA/specula/simul.py", line 507, in run
    self.loop.run(run_time=params['main']['total_time'], dt=params['main']['time_step'], speed_report=True)
  File "/Users/guido/GitHub/SPECULA/specula/loop_control.py", line 64, in run
    self.iter()
  File "/Users/guido/GitHub/SPECULA/specula/loop_control.py", line 110, in iter
    element.trigger()
  File "/Users/guido/GitHub/SPECULA/specula/base_processing_obj.py", line 155, in trigger
    self.trigger_code()
  File "/Users/guido/GitHub/SPECULA/specula/processing_objects/atmo_evolution.py", line 273, in trigger_code
    layer_phase = (1.0 - new_position_rem[ii]) * p[0: ipli, pos: ipli_p] + new_position_rem[ii] * p[0: ipli, pos+1: ipli_p+1]
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ValueError: operands could not be broadcast together with shapes (160,160) (160,159)
```

alfiopuglisi:
Fixed by #234

PR #234, opened 2025-07-31
Atmo bug fix

alfiopuglisi:
Fixes issue #80 and other bugs


specula/processing_objects/atmo_evolution.py
```diff
@@ -144,113 +139,91 @@ def compute(self):
...
+            # Number of rectangular phase screens from a single square phasescreen
+            n_ps_from_square_ps = self.xp.floor(self.pixel_square_phasescreens / self.pixel_phasescreens)
+            # Number of square phasescreens
+            n_ps = self.xp.ceil(float(self.n_phasescreens) / n_ps_from_square_ps)
+
+            # Seed vector
+            seed = self.xp.arange(self.seed, self.seed + int(n_ps))
+
+            # Square phasescreens
+            if self.make_cycle:
```
guidoagapito: The option self.make_cycle is no more required because we have the atmo_infinite_evolution that was not present in PASSATA.

specula/processing_objects/atmo_evolution.py
```diff
@@ -144,113 +139,91 @@ def compute(self):
...
 
-                    ps_index += 1
+            temp_screen = self.to_xp(temp_screen, dtype=self.dtype)
+            temp_screen *= self.xp.sqrt(self.Cn2[i])
+            temp_screen -= self.xp.mean(temp_screen)
 
-                    self.phasescreens.append(temp_screen)
-                    self.phasescreens_sizes.append(temp_screen.shape[1])
+            # Convert to nm
+            temp_screen *= self.wavelengthInNm / (2 * np.pi)
```
guidoagapito: self.wavelengthInNm must be 500 to get the correct conversion, we could hardcoded it instead of having an argument with a default value.

guidoagapito (approved):

specula/processing_objects/atmo_evolution.py
FabioRossiArcetri: We should either remove it from the init params list or implement it if it could be useful for comparison with the infinite atmo evolution.

specula/processing_objects/atmo_evolution.py
FabioRossiArcetri: I agree, there could be a constant defined in this file, and the constant could be also the default for wavelengthInNm, and in the code the constant might be used.

specula/processing_objects/atmo_infinite_evolution.py
FabioRossiArcetri, line 1: I would like to keed the l0 parameter even if it is not used now could be used in the future

alfiopuglisi: Ok the l0 changes have been reverted

specula/processing_objects/atmo_evolution.py
alfiopuglisi: I removed the make_cycle parameter and related code, which was commented-out anyway

FabioRossiArcetri (approved):

specula/processing_objects/atmo_evolution.py
alfiopuglisi: In another part of the code (trigger_code) it is used like this, scaling whatever wavelength was passed to 500

`        r0wavelength = r0 * (self.wavelengthInNm / 500.0)**(6./5.)`

maybe we can do the same here? Or is this r0 scaling a mistake?

I also checked PASSATA code and everything is identical, that is, this mistake has been "imported" from PASSATA as-is. Was the wavelength parameter ever used in AtmoEvolution?

guidoagapito: This "mistake" come from an old IDL code where the phasescreens unit was radians and not nm. Now wavelengthInNm must be always 500 and r0wavelength is equal to r0.

alfiopuglisi: Ok, I added a constant and removed the r0 conversion

guidoagapito (approved):

PR #242, opened 2025-08-03
Fixed warning on seeing vector size, removed unused lines

alfiopuglisi:
Deprecation warnings were generated from pytest.
Now it uses value[0] to extract a single array element as instructed by the deprecation warning

### DIFF
--- a/specula/processing_objects/atmo_evolution.py
+++ b/specula/processing_objects/atmo_evolution.py
@@ -26,6 +26,7 @@ class AtmoEvolution(BaseProcessingObj):
                  fov: float=0.0,
                  pixel_phasescreens: int=8192,
                  seed: int=1,
+                 extra_delta_time: float=0,
                  verbose: bool=False,
                  fov_in_m: float=None,
                  pupil_position:list =[0,0],
@@ -45,6 +46,7 @@ class AtmoEvolution(BaseProcessingObj):
         self.last_t = 0
         self.cycle_screens = True
         self.delta_time = None
+        self.extra_delta_time = extra_delta_time  # TODO to be used for speed-of-light corrections
                 
         self.inputs['seeing'] = InputValue(type=BaseValue)
         self.inputs['wind_speed'] = InputValue(type=BaseValue)
--- a/specula/processing_objects/atmo_evolution.py
+++ b/specula/processing_objects/atmo_evolution.py
@@ -43,13 +43,8 @@ class AtmoEvolution(BaseProcessingObj):
         self.n_phasescreens = len(heights)
         self.last_position = np.zeros(self.n_phasescreens)
         self.last_t = 0
-        self.extra_delta_time = 0
         self.cycle_screens = True
-
-        self.delta_time = 1
-        self.seeing = 1
-        self.wind_speed = 1
-        self.wind_direction = 1        
+        self.delta_time = None
                 
         self.inputs['seeing'] = InputValue(type=BaseValue)
         self.inputs['wind_speed'] = InputValue(type=BaseValue)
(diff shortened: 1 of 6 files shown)


specula/processing_objects/atmo_evolution.py
```diff
@@ -43,13 +43,8 @@ def __init__(self,
         self.n_phasescreens = len(heights)
         self.last_position = np.zeros(self.n_phasescreens)
         self.last_t = 0
-        self.extra_delta_time = 0
```
guidoagapito, line 46: In PASSATA we used self.extra_delta_time when it is important not to neglect the speed of light. For example in monostatic sodium laser system with pyramid where we want to estimate tip/tilt. I think we should keep this parameter and add it to the init arguments.

FabioRossiArcetri: @alfiopuglisi : can you add it back?

FabioRossiArcetri: with a comment describing its meaning

FabioRossiArcetri (changes requested):
add back self.extra_delta_time = 0 with a description

specula/processing_objects/atmo_evolution.py
alfiopuglisi: Ok I added it also to the init arguments, however there is no implementation yet, so I left a TODO. Should we add it to AtmoInfiniteEvolution as well?

specula/processing_objects/atmo_evolution.py
```diff
@@ -214,15 +208,28 @@ def compute(self):
...
+        if len(self.local_inputs['wind_speed'].value) != self.n_phasescreens:
+            raise ValueError('Wind speed input must be a {self.n_phasescreens}-elements array')
+        if len(self.local_inputs['wind_direction'].value) != self.n_phasescreens:
+            raise ValueError('Wind direction input must be a {self.n_phasescreens}-elements array')
+        
     def prepare_trigger(self, t):
         super().prepare_trigger(t)
-        self.delta_time = self.t_to_seconds(self.current_time - self.last_t) + self.extra_delta_time        
-    
+        self.delta_time = self.t_to_seconds(self.current_time - self.last_t)
```
guidoagapito: This should be added here:

+ self.extra_delta_time

specula/processing_objects/atmo_infinite_evolution.py
```diff
@@ -347,14 +342,25 @@ def initScreens(self, seed):
...
+        
+        # Check that wind speed and direction have the correct length
+        if len(self.local_inputs['wind_speed'].value) != self.n_infinite_phasescreens:
+            raise ValueError('Wind speed input must be a {self.n_infinite_phasescreens}-elements array')
+        if len(self.local_inputs['wind_direction'].value) != self.n_infinite_phasescreens:
+            raise ValueError('Wind direction input must be a {self.n_infinite_phasescreens}-elements array')
 
     def prepare_trigger(self, t):
         super().prepare_trigger(t)
         self.delta_time = self.t_to_seconds(self.current_time - self.last_t)
```
guidoagapito, line 362: As for AtmoEvolution, this should be added here:

+self.extra_delta_time

alfiopuglisi: ok done

specula/processing_objects/atmo_evolution.py
alfiopuglisi: ok done

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #244, opened 2025-08-04
(GA) Extended source for pyramid WFS

guidoagapito:


specula/processing_objects/modulated_pyramid.py
```diff
@@ -94,6 +97,13 @@ def __init__(self,
         self.xShiftPhInPixel = xShiftPhInPixel
         self.yShiftPhInPixel = yShiftPhInPixel
         self.pup_shifts = pup_shifts
+        self.extended_source = extended_source
+        if self.extended_source is not None:
+            self.extended_source_in_on = True
```
alfiopuglisi: There is no need for a separate` extended_source_in_on` flag: checking whether `self.extended_source` is not None should be enough

specula/processing_objects/extended_source.py
```diff
@@ -0,0 +1,751 @@
+import numpy as np
+from typing import Optional, Union, Tuple, List
+from specula.base_processing_obj import BaseProcessingObj
+from specula.data_objects.simul_params import SimulParams
+from specula.base_value import BaseValue
+from specula.connections import InputValue
+
+class ExtendedSource(BaseProcessingObj):
```
alfiopuglisi, line 8: If ExtendedSource is a processing object, because the PSF might change, then:
- psf should be an optional input, not an init parameter. The compute method should recompute only if the psf has been refreshed
- the extended object must be an optional input to the pyramid, not a ref parameter

specula/processing_objects/modulated_pyramid.py
guidoagapito: Done

specula/processing_objects/extended_source.py
guidoagapito: PSF is an optional input, and the optional argument 'initial_psf' is required to obtain extended source data at the first iteration of the simulation.

The 'extended_object coeff' output becomes an optional input for the pyramid.

alfiopuglisi (approved):

PR #246, opened 2025-08-04
(GA) gain ramp in ANDES parameters

guidoagapito:


main/scao/params_andes.yml
```diff
@@ -128,18 +128,32 @@ modalrec:
...
+      [3.0,  1.0,  0.0,  0.0,  0.0,  0.0,  0.0,  0.0],
+      [3.0,  1.0, 1.0,  0.0,  0.0,  0.0,  0.0,  0.0], 
+      [2.5,  1.0, 1.0,  1.0,  0.0,  0.0,  0.0,  0.0],
+      [2.5,  1.0, 1.0,  1.0,  1.0,  0.0,  0.0,  0.0],
+      [2.0,  1.0,  1.0,  1.0,  1.0,  1.0,  0.0,  0.0],
+      [1.5,  1.0,  1.0,  1.0,  1.0,  1.0,  1.0,  0.0],
+      [1.5,  1.0,  1.0,  1.0,  1.0,  1.0,  1.0,  1.0],
+      [1.0,  1.0,  1.0,  1.0,  1.0,  1.0,  1.0,  1.0]
+  ]
+  time_intervals: [0.05,0.1,0.15,0.2,0.25,0.3,0.35]
```
alfiopuglisi: I think "intervals" is misleading, because it would mean that the first set is held for 0.05s, the second set for 0.1s, etc. Instead it's more similar to a "scheduled_times" list: the list of times at which each value is applied.

guidoagapito: Done, now it is scheduled_times.

alfiopuglisi (approved):

PR #251, opened 2025-08-05
(GA) FuncGenerator has been divided into several different generator classes

guidoagapito:


specula/processing_objects/base_generator.py
```diff
@@ -0,0 +1,66 @@
+from abc import ABC, abstractmethod
+
+from specula.base_value import BaseValue
+from specula.base_processing_obj import BaseProcessingObj
+
+
+class BaseGenerator(BaseProcessingObj, ABC):
```
alfiopuglisi: We had decided not to use ABC, since this is not a base class, it's already derived from BaseProcessingObj. It works now, but it introduces costrains on the base classes for the future.
I think it's enough to use a normal class here and, in the trigger_code method, raise a NotImplementedError so that a derived class must reimplement it.

specula/processing_objects/base_generator.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Base class for function generators.
+    
+    All specific generators inherit from this class and implement trigger_code().
+    """
+    def __init__(self,
+                 output_size: int = 1,
+                 constant: float = 0.0,
```
alfiopuglisi: Why the "constant" parameter in the base class? It's not used here and also not by some of the derived classes, it should be handled just in those derived classes that use it.

specula/processing_objects/push_pull_generator.py
```diff
@@ -0,0 +1,65 @@
...
+        if nsamples != 1 and push_pull_type != 'PUSHPULL':
+            raise ValueError('nsamples can only be used with PUSHPULL type')
+        
+        super().__init__(
+            output_size=nmodes,
+            target_device_idx=target_device_idx,
+            precision=precision
+        )
+        
+        self.nmodes = nmodes
```
alfiopuglisi: self.nmodes is not used

specula/processing_objects/vibration_generator.py
```diff
@@ -0,0 +1,94 @@
...
+                 precision: int = None):
+        
+        super().__init__(
+            output_size=nmodes,
+            target_device_idx=target_device_idx,
+            precision=precision
+        )
+        
+        self.simul_params = simul_params
+        self.nmodes = nmodes
```
alfiopuglisi: self.simul_params and self.nmodes are not used

specula/processing_objects/base_generator.py
guidoagapito: Done

specula/processing_objects/push_pull_generator.py
guidoagapito: Done

specula/processing_objects/vibration_generator.py
guidoagapito: Done

specula/processing_objects/base_generator.py
guidoagapito: Done

alfiopuglisi (approved):

PR #272, opened 2025-08-19
Simul params

alfiopuglisi:
SimulParams now derives from BaseDataObj


specula/simul.py
```diff
@@ -688,7 +687,6 @@ def run(self):
 
         # Actual creation code
         self.apply_overrides(params)
-        self.setSimulParams(params)
```
guidoagapito, line 691: setSimulParams method is still there, Is it desired?

FabioRossiArcetri: Yes nothing changed about this. The simulation has its simulParams member, which is set when the obects are created and after the overrides are applied.

FabioRossiArcetri (approved):

guidoagapito (approved):

PR #273, opened 2025-08-19
PSF geometry calculation

alfiopuglisi:
This PR factors out the PSF geometry calculation into a library function, since it is used in multiple places.


test/test_psf geometry.py
guidoagapito, line 1: I think we should use a name without space but with an underscore.

specula/processing_objects/psf.py
```diff
@@ -78,75 +66,6 @@ def setup(self):
...
-        
-        Returns:
-            pixel_size_mas 
-        """
-
-        pixel_size_mas = (self.wavelengthInNm * 1e-9 / self.dim_pup_in_m * 3600 * 180 / np.pi) * 1000 / self.nd
-
-        return pixel_size_mas
-
     def calc_psf(self, phase, amp, imwidth=None, normalize=False, nocenter=False):
```
guidoagapito: I think it could be useful to have a static method to compute PSF since we remove the function from the library.

test/test_psf geometry.py
alfiopuglisi: done, actually now there are two different files: test_psf.py and test_calc_psf.py

specula/processing_objects/psf.py
alfiopuglisi: Done, all the calculations moved to lib/calc_psf.py

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #286, opened 2025-08-21
(GA) mod_type and test

guidoagapito:


specula/processing_objects/modulated_pyramid.py
```diff
@@ -384,16 +406,62 @@ def cache_ttexp(self):
...
+                        else:
+                            # Middle points: average over symmetric interval
+                            avg_tilt = self.mod_amp * (2 * tt / (self.mod_steps - 1) - 1)
+
+                        # Convert to flux factor - INVERTED: more weight at the edges
+                        normalized_angle = self.xp.abs(avg_tilt) * self.xp.pi / (2 * self.mod_amp)
+                        # Use 1/cos(angle) to compensate for intensity loss at large tilts
+                        # Add small epsilon to avoid division by zero
+                        epsilon = 1e-8
+                        self.flux_factor_vector[tt] = 1.0 / (self.xp.cos(normalized_angle) + epsilon)
```
alfiopuglisi: This means that we expect flux factors equal to 1e8 ?

specula/processing_objects/modulated_pyramid.py
```diff
@@ -457,8 +525,12 @@ def prepare_trigger(self, t):
         self._wf_interpolated.ef_at_lambda(self.wavelength_in_nm, out=self.ef)
 
     def trigger_code(self):
+        # Rotate ttexp for alternating modulation, for standard modulation self.rotation_k = 0
+        # self.ttexp is 3D and we rotate on the last two axes
+        ttexp_rotated = self.xp.rot90(self.ttexp, k=self.rotation_k[self.iter % 2], axes=(-2, -1))
```
alfiopuglisi: I think that this will fail when using streams: self.iter%2 changes at each call of trigger_code(), but stream capture will only get the first value. So all calculations will be done with self.rotation_k[0]

alfiopuglisi: maybe it will work if self.iter is a cupy array, but a test is needed...

specula/processing_objects/modulated_pyramid.py
```diff
@@ -474,6 +546,7 @@ def trigger_code(self):
         self.psf_bfm_arr[:] = self.xp.fft.fftshift(self.fpsf)
         self.psf_tot_arr[:] = self.psf_bfm_arr * self.fp_mask
         self.pup_pyr_tot[:] = self.xp.roll(self.pyr_image, self.roll_array, self.roll_axis )
+        print('sum of pup_pyr_tot:', self.xp.sum(self.pup_pyr_tot))
```
alfiopuglisi: This print will not be executed when stream capture is done, or worse, it will be executed during capture printing incorrect values because there is no valid input yet.

guidoagapito: Done

specula/processing_objects/modulated_pyramid.py
guidoagapito: self.to_xp used to make an array of 1 element

specula/processing_objects/modulated_pyramid.py
guidoagapito: removed

alfiopuglisi (approved):

FabioRossiArcetri (approved):

PR #288, opened 2025-08-21
Small optimizations

alfiopuglisi:
DtD, DtH etc have been moved inside the "if" clause, so that they are not executed in the majority of cases.


guidoagapito (approved):

specula/base_time_obj.py
```diff
@@ -71,19 +71,7 @@ def t_to_seconds(self, t):
...
-        if '.' not in ss:
-            ss += '.0'
-
-        dotpos = ss.find('.')
-        intpart = ss[:dotpos]
-        fracpart = ss[dotpos + 1:]
-
-        return (int(intpart) * self._time_resolution +
-                int(fracpart) * (self._time_resolution // (10 ** len(fracpart))))
+        return int(round(seconds, ndigits=9) * self._time_resolution)
```
FabioRossiArcetri, line 74: Are you sure the round is to be performed before the multiplication by the time resolution?

FabioRossiArcetri (commented):
Check the comment

specula/base_time_obj.py
guidoagapito: I think it is ok, but the number of digits should be a function of self._time_resolution

FabioRossiArcetri: ok, 1 nanosecond seams enough.

PR #294, opened 2025-08-21
(GA) Doc update

guidoagapito:


docs/scripts/generate_api_docs.py
```diff
@@ -0,0 +1,135 @@
+import os
```
alfiopuglisi, line 1: Why a separate file? Sphinx auto-discovery of python modules and classes doesn't work?

guidoagapito: I added this to have more flexibility in the organization of the the doc and to avoid one page for file, but if you prefer I can add a sphinx-apidoc command in the make file.

alfiopuglisi: No it's ok, we can keep it as it is, I was just wandering.

alfiopuglisi (approved):

PR #297, opened 2025-08-21
(GA) poly chromatic SH with test

guidoagapito:


guidoagapito:
The test was also run on gandalf.

FabioRossiArcetri (approved):

specula/processing_objects/poly_chrom_sh.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def __init__(self,
+                 wavelengthInNm: list,           # List of wavelengths for each SH
+                 flux_factor: list,                 # Flux factor for each wavelength
+                 xy_tilts_in_arcsec: list = None,   # Optional differential tilts [dx, dy] for each SH
+                 # SH parameters (shared by all SH instances)
+                 subap_wanted_fov: float = 1.0,
+                 sensor_pxscale: float = 1.0,
+                 subap_on_diameter: int = 1,
+                 subap_npx: int = 10,
```
alfiopuglisi: These default values for subap_wanted_fov, pxscale, etc (the last four parameters) look like testing values that will never be used in real simulations. Maybe it's better to make these parameters non-optional?

specula/processing_objects/poly_chrom_sh.py
```diff
@@ -0,0 +1,225 @@
...
+                 **sh_kwargs                        # Additional SH parameters
```
alfiopuglisi: The SH class does not have so many more parameters that the ones listed. Can we just add them explicitly, avoiding the **kwargs? In this way, it is a lot easier to know which are the required parameters for this class, instead of looking up a different one

alfiopuglisi:
Apart from the previous comments, this class looks very similar to the DistributedSH one, just with more features (local tilts, different wavelengths and fluxes), and using "whole" SHs to be summed up instead of partial SHs to be concatenated.

I would try to make a single class to implement both functionalities.

specula/processing_objects/poly_chrom_sh.py
guidoagapito: Done

specula/processing_objects/poly_chrom_sh.py
guidoagapito: Done

alfiopuglisi (approved):

PR #298, opened 2025-08-22
(GA) max_stable_gain with test and fix to gain optimizer

guidoagapito:


specula/processing_objects/gain_optimizer.py
```diff
@@ -166,16 +182,20 @@ def _optimize_gains(self, t):
             print(f"Optimized gains at t={self.t_to_seconds(t):.3f}s: "
                   f"mean={float(self.xp.mean(opt_gains)):.4f}")
 
-        # Clear history
-        self._clear_history()
+        # Optionally clear cache periodically to avoid memory buildup
```
alfiopuglisi: Since the lru_cache has a maxsize parameter, is this clearing really necessary? It should be automatic. If too much memory is used, then we should reduce the maxsize value.

specula/processing_objects/gain_optimizer.py
```diff
@@ -302,42 +342,121 @@ def _calculate_psd(self, data, t_int):
...
+        return freq_rounded, t_int_rounded, gain_rounded, num_rounded, den_rounded, delay_rounded
+
+    @lru_cache(maxsize=2048)
+    def _calculate_rejection_tf_cached(self, freq_tuple, t_int, gain, num_tuple, den_tuple, delay):
+        """
+        Calculate rejection transfer function using lru_cache.
+        All parameters must be hashable (tuples, not arrays).
+        """
+        # Convert tuples back to arrays
+        freq = self.to_xp(np.array(freq_tuple), dtype=self.dtype)
```
alfiopuglisi: I think that np.array(x) is not necessary when using to_xp(), the conversion from a tuple to array should be automatic

specula/processing_objects/gain_optimizer.py
```diff
@@ -302,42 +342,121 @@ def _calculate_psd(self, data, t_int):
...
 
-        # Clear history lists
-        self.time_hist = []
-        self.delta_comm_hist = []
-        self.comm_hist = []
+        except Exception as e:
+            # Fallback to direct calculation if cache fails
+            self._cache_misses += 1
+            if self.verbose:
+                print(f"Cache miss, calculating directly: {e}")
```
alfiopuglisi: This does not look right: `_calculate_rejection_tf_cached()` will not raise any exception if there is a cache miss, it will just recalculate things.
Also the "Direct calculation" in this except clause is a duplication of the code in that function.
Finally, if we really need an exception, we should not catch just Exception but a specific type. Catching Exception is reserved for last-resort error handling

specula/data_objects/iir_filter_data.py
```diff
@@ -402,6 +412,328 @@ def is_stable(self, mode, dm=None, nw=None, dw=None, verbose=False):
...
+        in a closed-loop system with the given plant dynamics.
+        
+        Args:
+            mode: Filter mode index. If None, calculates for all modes
+            delay: Delay in frames (alternative to dm/nw/dw)
+            dm, nw, dw: Plant coefficients (alternative to delay)
+            max_gain: Maximum gain to test (default: 20.0)
+            n_gain: Number of gain values to test (default: 10000)
+            tolerance: Minimum gain to test (default: 1e-6)
+            use_cache: Whether to use caching for identical filters (default: True)
```
alfiopuglisi: Is there a reason not to use the cache?

FabioRossiArcetri (approved):

specula/data_objects/iir_filter_data.py
guidoagapito: No, we can usa a cache here.

alfiopuglisi (approved):

PR #310, opened 2025-08-24
(GA) Generic poly chromatic WFS and Pyramid with test

guidoagapito:


specula/processing_objects/poly_chrom_wfs.py
```diff
@@ -0,0 +1,202 @@
...
+        if 0 <= index < self.n_wavelengths:
+            wfs = self._wfs_instances[index]
+            flux_factor = self.flux_factor_normalized[index]
+            return wfs.outputs['out_i'].i * flux_factor
+        else:
+            raise IndexError(f"Wavelength index {index} out of range [0, {self.n_wavelengths-1}]")
+
+    # This method must be implemented in subclasses!
+    # It exists to make this base class abstract
+    def _check_subwfs(self):
```
alfiopuglisi: This method can be removed from all classes, I don't find it very useful since subWFSs are created by the same class that should check them.

specula/processing_objects/poly_chrom_wfs.py
```diff
@@ -0,0 +1,202 @@
...
+                self._out_i.i *= total_input_flux / current_total
+
+    def get_wfs_instance(self, index):
+        """Get a specific WFS instance for debugging or analysis."""
+        if 0 <= index < self.n_wavelengths:
+            return self._wfs_instances[index]
+        else:
+            raise IndexError(f"WFS index {index} out of range [0, {self.n_wavelengths-1}]")
+
+    def get_wavelength_contribution(self, index):
```
alfiopuglisi, line 175: This method is not very useful (not called anywhere except for tests), unless we want to create an additional output with the intensity contribution of each WFS

specula/processing_objects/poly_chrom_wfs.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Optional: normalize total intensity to match input photon flux
+        in_ef = self.local_inputs['in_ef']
+        if hasattr(in_ef, 'S0') and in_ef.S0 > 0:
+            total_input_flux = in_ef.S0 * in_ef.masked_area()
+            current_total = self.xp.sum(self._out_i.i)
+            if current_total > 0:
+                self._out_i.i *= total_input_flux / current_total
+
+    def get_wfs_instance(self, index):
```
alfiopuglisi: This method is not used anywhere, can be removed

specula/processing_objects/poly_chrom_wfs.py
```diff
@@ -0,0 +1,202 @@
...
+            # Add weighted contribution
+            flux_factor = self.flux_factor_normalized[i]
+            self._out_i.i += wfs.outputs['out_i'].i * flux_factor
+
+        # Set generation time
+        self._out_i.generation_time = self.current_time
+
+        # Optional: normalize total intensity to match input photon flux
+        in_ef = self.local_inputs['in_ef']
+        if hasattr(in_ef, 'S0') and in_ef.S0 > 0:
```
alfiopuglisi: in_ef should always have an S0 attribute, so this check can be removed

specula/processing_objects/poly_chrom_wfs.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Prepare all SH instances
+        for wfs in self._wfs_instances:
+            wfs.prepare_trigger(t)
+
+    def trigger_code(self):
+        # Reset output intensity
+        self._out_i.i[:] = 0.0
+
+        # Trigger each SH and accumulate results
```
alfiopuglisi: Remove "and accumulate results", it is not done here.

specula/processing_objects/poly_chrom_wfs.py
```diff
@@ -0,0 +1,202 @@
...
+        # Update all modified EFs
+        for i, modified_ef in enumerate(self._modified_efs):
+            # Copy basic properties
+            modified_ef.A[:] = in_ef.A
+            modified_ef.S0 = in_ef.S0
+            modified_ef.pixel_pitch = in_ef.pixel_pitch
+
+            # Apply tilt if present
+            tilt_x, tilt_y = self.xy_tilts_in_arcsec[i]
+            if tilt_x != 0.0 or tilt_y != 0.0:
```
alfiopuglisi: I would simplify the code removing this if and keeping the full computation: if tilts are zero the results are correct anyway, and the amount of computation is trivial compared to the SH code.
This also means to remove the self._has_tilts flag, and always compute them (which is better because now it creates additional self members depending on the flag, which is not really correct)

guidoagapito:
I agree with the proposed changes. You can either wait for me to return, or proceed with them yourself. Thanks!

alfiopuglisi:
> I agree with the proposed changes. You can either wait for me to return, or proceed with them yourself. Thanks!

I will implement them and merge

PR #339, opened 2025-09-05
Better diagrams

FabioRossiArcetri:


alfiopuglisi:
Looks good to me

specula/simul.py
```diff
@@ -1,4 +1,5 @@
 import re
+from tkinter.font import BOLD
```
alfiopuglisi, line 2: Is this import from tkinter.font needed ?

alfiopuglisi: Is this import from tkinter.font needed ?

alfiopuglisi (approved):

PR #349, opened 2025-09-09
Minor cleanup

alfiopuglisi:
Fixed linting errors, and removed some unused or duplicated code


specula/base_time_obj.py
```diff
@@ -45,7 +44,7 @@ def __init__(self, target_device_idx=None, precision=None):
             self.xp_str = 'np'
 
         if self.target_device_idx>=0:
-            from cupyx.scipy.ndimage import rotate
+            from cupyx.scipy.ndimage import rotate as ndimage_rotate
             from cupyx.scipy.ndimage import shift
```
guidoagapito: the same could be done for shift --> ndimage_shift

alfiopuglisi: I found out that this shift routine is only used by InfinitePhaseScreen, and it is commented out, so maybe it can be removed. @FabioRossiArcetri what do you think?

guidoagapito: I think it could be useful in future.

guidoagapito (approved):

FabioRossiArcetri (approved):

specula/base_time_obj.py
FabioRossiArcetri: I would say if these are not used we can remove for now.

PR #350, opened 2025-09-09
Generators can stop simulation when done

alfiopuglisi:
Allow push pull and time history generators to stop the simulation when done, so that calibration yml files do not need to calculate the number of iterations for e.g. push-pull history.


alfiopuglisi:
Does not work because main.total_time=0.010 and since the override is removed, the simulation ends before the generator has a chance of raising the stop exception.

test/params_scao_sh_test_rec.yml
```diff
@@ -7,9 +7,6 @@ pushpull:
                     50,50,50,50,50,50,50,50,50,50]
   outputs: ['output']
 
-main_override:
-  total_time:  0.080
```
guidoagapito, line 11: Could you put a time larger than the requested one to test this new feature?

alfiopuglisi: Yes and it would work, but then one has to add the override in any case, I would like to find a way to avoid the total_time override at all.

guidoagapito: Could we remove the total_time key with the override? For clarity reasons in the yml file I would like to avoid having a parameter that is ignored.

FabioRossiArcetri:
Can we can think about different ways to implement this feature? In my view a generator should not have this 'superpower'. Maybe a calibrator could?

guidoagapito:
> Can we can think about different ways to implement this feature? In my view a generator should not have this 'superpower'. Maybe a calibrator could?

For me the negative point is that this super-power is hidden and not visible in the yml file.

alfiopuglisi:
To make it more visible, total_time could be set to 'auto' or null.
The reason for this feature is that sometimes it is difficult to calculate the correct length, you have to find out the push-pull length and combine it with the simulation time step and it is easy to make mistakes.

guidoagapito:
> To make it more visible, total_time could be set to 'auto' or null. The reason for this feature is that sometimes it is difficult to calculate the correct length, you have to find out the push-pull length and combine it with the simulation time step and it is easy to make mistakes.

I think it is a reasonable approach.

PR #353, opened 2025-09-10
Folder names

FabioRossiArcetri:


alfiopuglisi:
I realize now that many yaml files have path like /raid1/guido or /raid2/gcarla, which should not really be committed into the main SPECULA repo. We need to think of a mechanism for user config files, but after this PR

alfiopuglisi (approved):

guidoagapito (approved):

PR #359, opened 2025-09-12
(GA) updates for running MORFEO simulations with platescale correction, sodium beam and slice of commands for each DM

guidoagapito:


config/MORFEO/params_morfeo_no_focus_no_ref.yml
```diff
@@ -0,0 +1,531 @@
...
+  laser_launch_tel_ref: 'launcher1'
+  inputs:
+    in_ef: 'prop.out_source_lgs1_ef'
+    sodium_altitude: 'sodium_altitude.output'
+    sodium_intensity: 'sodium_intensity.output'
+  outputs:  ['out_i']
+
+sh_lgs2: { <<: *LGS_SH, rotAnglePhInDeg: 14.2, laser_launch_tel_ref: 'launcher1',
+          inputs: {in_ef: 'prop.out_source_lgs2_ef', 
+                   sodium_altitude: 'sodium_altitude.output', sodium_intensity: 'sodium_intensity.output'}}
```
alfiopuglisi: These sodium parameters are identical to sh_lgs1 and do not need to be repeated

config/MORFEO/params_morfeo_no_focus_no_ref.yml
```diff
@@ -0,0 +1,531 @@
...
+
+dm1:
+  class:             'DM'
+  simul_params_ref:  'main'
+  ifunc_ref:         'ifunc_m4'
+  m2c_ref:           'm2c_m4'
+  nmodes:            4519
+  height:            600                      # DM height [m]
+  inputs:
+    in_command: 'slicer1.out_value'
```
alfiopuglisi, line 382: The DM class has an "input_offset" parameter that, together with the number of modes, acts as a slice on the input commands (see here: https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/0d92e6eb30eafe874773ad2e0f781fbaef90bb57/specula/processing_objects/dm.py#L111 ) .
Is it really necessary here to use an extra BaseSlicer object ? I think the functionality would be the same.

We can keep the BaseSlicer class anyway for general use.

alfiopuglisi: Or maybe we can rewrite the DM parameters to be the same as Modalrec, for consistency: add an "input_modes_slice" parameter to slice the input commands, removing the nmodes and input_offset parameters. We keep the start_mode because it is an offset into the ifunc (maybe renaming to something clearer like "first_ifunc_mode").

config/MORFEO/params_morfeo_no_focus_no_ref.yml
guidoagapito: If the inputs section is updated, all keys and values must be repeated.
YAML does not perform a deep merge.

config/MORFEO/params_morfeo_no_focus_no_ref.yml
guidoagapito: I did this because:

- Compatibility with FieldAnalyser: a specific data object for each DM.
- Clarity: I made the slicing operation explicit instead of hiding it inside the DM.
- Matching the behaviour of real systems: (in the systems I know) DM only receives its own commands.

config/MORFEO/params_morfeo_no_focus_no_ref.yml
guidoagapito: Done, I used a &LGS_SH_INPUTS for the inputs.

FabioRossiArcetri (approved):

alfiopuglisi (approved):

PR #364, opened 2025-09-15
Reactivating Intmat test on Windows

alfiopuglisi:
Test PR to check Windows tests

### DIFF
--- a/specula/data_objects/intmat.py
+++ b/specula/data_objects/intmat.py
@@ -157,6 +157,7 @@ class Intmat(BaseDataObj):
         # Reading additional fits extensions
         with fits.open(filename) as hdul:
             num_ext = len(hdul)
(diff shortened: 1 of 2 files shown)


specula/data_objects/intmat.py
```diff
@@ -157,6 +157,7 @@ def restore(filename, target_device_idx=None):
         # Reading additional fits extensions
         with fits.open(filename) as hdul:
             num_ext = len(hdul)
+        hdul.close() # Force close for Windows
```
guidoagapito: In other data objects and tests we used .copy() to avoid this kind of issue with Windows, see for example https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/specula/data_objects/pupilstop.py

PR #366, opened 2025-09-15
(GA) Focal plane filter with tests

guidoagapito:


alfiopuglisi:
It looks like a pyramid without the actual pyramid - just the focal plane mask. Is it correct?

FabioRossiArcetri (approved):

guidoagapito:
> It looks like a pyramid without the actual pyramid - just the focal plane mask. Is it correct?

Yes. I consider inheriting the pyramid or using it as base class for the pyramid, but it seems to me that there is no bug advantage. What do you think?

alfiopuglisi:
> > It looks like a pyramid without the actual pyramid - just the focal plane mask. Is it correct?
> 
> Yes. I consider inheriting the pyramid or using it as base class for the pyramid, but it seems to me that there is no bug advantage. What do you think?

Actually I was thinking that there is a quite a bit of complicated code for calc_geometry, interpolation etc. so it could be derived from the pyramid, overriding trigger(), or maybe even leaving all the pyramid as it is and just have a special initialization with mod step=1, mod amp=0, and a zero (or unitary?) pyramid EF (the self.shifted_masked_exp). It could be worth to do an experiment

guidoagapito:
> > > It looks like a pyramid without the actual pyramid - just the focal plane mask. Is it correct?
> > 
> > 
> > Yes. I consider inheriting the pyramid or using it as base class for the pyramid, but it seems to me that there is no bug advantage. What do you think?
> 
> Actually I was thinking that there is a quite a bit of complicated code for calc_geometry, interpolation etc. so it could be derived from the pyramid, overriding trigger(), or maybe even leaving all the pyramid as it is and just have a special initialization with mod step=1, mod amp=0, and a zero (or unitary?) pyramid EF (the self.shifted_masked_exp). It could be worth to do an experiment

Could you do this experiment?

alfiopuglisi:
> > > > It looks like a pyramid without the actual pyramid - just the focal plane mask. Is it correct?
> > > 
> > > 
> > > Yes. I consider inheriting the pyramid or using it as base class for the pyramid, but it seems to me that there is no bug advantage. What do you think?
> > 
> > 
> > Actually I was thinking that there is a quite a bit of complicated code for calc_geometry, interpolation etc. so it could be derived from the pyramid, overriding trigger(), or maybe even leaving all the pyramid as it is and just have a special initialization with mod step=1, mod amp=0, and a zero (or unitary?) pyramid EF (the self.shifted_masked_exp). It could be worth to do an experiment
> 
> Could you do this experiment?

I can try tomorrow

alfiopuglisi (approved):

PR #390, opened 2025-09-19
Leonardo work and more

FabioRossiArcetri:


specula/simul.py
```diff
@@ -340,9 +340,9 @@ def build_objects(self, params):
             # If not build, remember the remote rank of this object (needed for connections setup)
             if not build_this_object:
                 self.remote_objs_ranks[key] = target_rank
-                continue
+                # continue
```
alfiopuglisi: We can remove this commented-out "continue" and maybe add a comment that we do go through the rest of the code anyway in order to build the connection diagram (if I understood correctly)

guidoagapito (approved):

specula/simul.py
FabioRossiArcetri: yes correct, I am removing the commented line.

alfiopuglisi (approved):

PR #403, opened 2025-10-02
Update of physical propagation

bernadett92:
Updated/implemented the physical propagation

### DIFF
--- a/specula/processing_objects/atmo_propagation.py
+++ b/specula/processing_objects/atmo_propagation.py
@@ -1,4 +1,3 @@
-
 from specula.lib.make_xy import make_xy
 from specula.lib.utils import local_mean_rebin
 from specula.base_processing_obj import BaseProcessingObj
@@ -8,6 +7,9 @@ from specula.connections import InputList
 from specula.data_objects.layer import Layer
 from specula import cpuArray, show_in_profiler
 from specula.data_objects.simul_params import SimulParams
+from skimage.filters import window
+from symao.turbolence import ft_ft2
+from symao.turbolence import ft_ift2
 
 import numpy as np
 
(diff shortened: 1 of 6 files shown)


specula/processing_objects/atmo_propagation.py
```diff
@@ -8,6 +7,9 @@
 from specula.data_objects.layer import Layer
 from specula import cpuArray, show_in_profiler
 from specula.data_objects.simul_params import SimulParams
+from skimage.filters import window
```
guidoagapito, line 10: skimage should be added to the specula requirements, see https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/requirements.txt
@FabioRossiArcetri what do you think?

guidoagapito: Test is failing because skimage is not in the requirements.

guidoagapito: Done

FabioRossiArcetri: It's ok, we can add it, it's a commmonly used lib.

FabioRossiArcetri (approved):

guidoagapito:
Please update https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/docs/installation.rst and https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/README.md with the info about new requirements.

guidoagapito:
Another request: please merge main to avoid issue with test_infinite_phase_screen.py solve recently.

guidoagapito (approved):

alfiopuglisi (approved):

specula/processing_objects/atmo_propagation.py
```diff
@@ -8,6 +7,9 @@
...
+from symao.turbolence import ft_ft2
+from symao.turbolence import ft_ift2
```
alfiopuglisi, line 12: Just a comment: i think that the correct english word is "turbulence" with an "u", however it is coming from another module and the change should be made in symao.
Also the two ft_ft2 and ft_ift2 functions are very simple and maybe there is not even a need to import them, but we can do another pass at this later on.

PR #404, opened 2025-10-02
(GA) bug in SH slopec with weights for WCoG and weights from average intensity

guidoagapito:


specula/processing_objects/sh_slopec.py
```diff
@@ -142,13 +145,13 @@ def computeXYweights(self, np_sub, exp_weight, weightedPixRad, quadcell_mode=Fal
         return {"x": x, "y": y, "xc": xc, "yc": yc, "mask_weighted": mask_weighted}
 
     def trigger_code(self):
-        if self.vecWeiPixRadT is not None:
+        if self.vec_wei_pix_rad_t is not None:
             time = self.current_time_seconds
-            idxW = self.xp.where(time > self.vecWeiPixRadT[:, 1])[-1]
+            idxW = self.xp.where(time > self.vec_wei_pix_rad_t[:, 1])[-1]
```
alfiopuglisi: if "time" is only used here, I would put "self.current_time_seconds" directly in the subsequent expression, so that we avoid shadowing the "time" module

guidoagapito: Done

specula/processing_objects/sh_slopec.py
```diff
@@ -194,32 +196,35 @@ def calc_slopes_nofor(self):
...
-                    # Count subapertures where weights were applied
                     n_weight_applied = self.xp.sum(self.xp.any(int_pixels_weight > 0, axis=0))
                 else:
                     # Normalize by max value for valid subapertures
                     int_pixels_weight[:, valid_mask] /= max_temp[valid_mask]
                     int_pixels_weight[:, ~valid_mask] = 1.0
                     n_weight_applied = self.xp.sum(valid_mask)
 
-                self.int_pixels_weight[:] = int_pixels_weight
+                self.int_pixels_weight[:] = int_pixels_weight.astype(self.dtype)
```
alfiopuglisi: Is this astype() necessary? It should automatically take the dtype of the destination array

specula/processing_objects/sh_slopec.py
```diff
@@ -276,15 +281,32 @@ def calc_slopes_nofor(self):
...
+        Args:
+            np_sub (int): Number of sub-apertures (pixels) in one dimension.
+            fwhm (list): Full width at half maximum (FWHM) in pixels for x and y directions.
+
+        Returns:
+            np.ndarray: 2D array representing the Gaussian PSF.
+        """
+        cntrd = (np_sub - 1) / 2.0
+
+        x = np.arange(np_sub) - cntrd  # da -(np_sub-1)/2 a +(np_sub-1)/2
```
alfiopuglisi: translate "da" "a" to "from" "to"

specula/processing_objects/sh_slopec.py
```diff
@@ -179,7 +181,7 @@ def calc_slopes_nofor(self):
         if self.weight_int_pixel:
 
             if self.int_pixels_weight is None:
-                self.int_pixels_weight = self.xp.ones_like(pixels)
+                self.int_pixels_weight = self.xp.ones_like(pixels, dtype=self.dtype)
```
alfiopuglisi, line 184: This comment applies to line 190, where I can't leave a comment:
be careful because self.int_pixels.pixels comes from a Pixels objects, which is by default defined as unsigned. Thus the subtraction might not have the intended effect (it could wrap around!). It is safe as it is, because it is subtracting the minimum value, but maybe a comment warning can be added. The "negative max values" in the comment at line 193 might not happen at all.
Also, depending on how smart the indexing in line 189 is, it might be modifying the input array with an in-place subtraction.

specula/processing_objects/sh_slopec.py
```diff
@@ -194,32 +196,35 @@ def calc_slopes_nofor(self):
...
+                    int_pixels_weight = window_mask.astype(self.dtype)
+
+                    # Handle invalid subapertures
+                    int_pixels_weight[:, ~valid_mask] = 1.0
 
-                    # Count subapertures where weights were applied
                     n_weight_applied = self.xp.sum(self.xp.any(int_pixels_weight > 0, axis=0))
                 else:
                     # Normalize by max value for valid subapertures
                     int_pixels_weight[:, valid_mask] /= max_temp[valid_mask]
```
alfiopuglisi, line 218: This looks like an integer division, since int_pixel_weight comes from the Pixels array. Is this what is wanted?

guidoagapito: Now int_pixels_weight is of self.dtype type

specula/processing_objects/sh_slopec.py
guidoagapito: I added astype(self.dtype)

specula/processing_objects/sh_slopec.py
guidoagapito: Done

specula/processing_objects/sh_slopec.py
guidoagapito: Removed

alfiopuglisi (approved):

PR #410, opened 2025-10-21
(GA) Flux per sub-aperture in SH and PYR slopec

guidoagapito:


specula/processing_objects/pyr_slopec.py
```diff
@@ -92,8 +92,14 @@ def trigger_code(self):
...
 
+        # Compute flux per subaperture (sum of all 4 pupils)
+        flux_per_subap = A + B + C + D
+        self.flux_per_subaperture_vector.value[:] = flux_per_subap
+
         # Compute total intensity
-        self.total_intensity = self.xp.sum(self.flat_pixels[self.pup_idx])
+        self.total_intensity = self.xp.sum(flux_per_subap)
+        self.total_counts.value[0] = self.total_intensity
+        self.subap_counts.value[0] = self.xp.mean(flux_per_subap)
```
alfiopuglisi: The assignments of self.total_counts.value[0] and self.subap_counts[0].value are overwring the ones made in line 84 and 85, which were performed before the clamp to zero. Which is the correct one?

specula/processing_objects/pyr_slopec.py
```diff
@@ -127,4 +133,6 @@ def post_trigger(self):
         super().post_trigger()
 
         self.outputs['out_pupdata'].generation_time = self.current_time
-
+        self.outputs['out_flux_per_subaperture'].generation_time = self.current_time
+        self.outputs['out_total_counts'].generation_time = self.current_time
+        self.outputs['out_subap_counts'].generation_time = self.current_time
```
alfiopuglisi: These same generation_time assignements are also made in the base Slopec class, so one of the two copies should be removed. Maybe it is actually better to keep them here in the derived class, since the output was calculated here.

specula/processing_objects/sh_slopec.py
```diff
@@ -310,4 +310,7 @@ def psf_gaussian(self, np_sub, fwhm):
 
     def post_trigger(self):
         super().post_trigger()
-        self.outputs['out_subapdata'].generation_time = self.current_time
\ No newline at end of file
+        self.outputs['out_subapdata'].generation_time = self.current_time
+        self.outputs['out_flux_per_subaperture'].generation_time = self.current_time
+        self.outputs['out_total_counts'].generation_time = self.current_time
+        self.outputs['out_subap_counts'].generation_time = self.current_time
```
alfiopuglisi: Same comment as PyrSlopec.post_trigger() for the generation time

test/test_sh_slopec.py
```diff
@@ -343,3 +343,53 @@ def test_shslopec_interleaved_slopesnull(self, target_device_idx, xp):
...
+
+        # Verify subap_counts is mean of flux_per_subaperture
+        expected_mean = xp.mean(flux_per_subap)
+        np.testing.assert_almost_equal(cpuArray(subap_counts[0]),
+                                       cpuArray(expected_mean), decimal=5)
+
+        # Verify all values are positive
+        self.assertTrue(xp.all(flux_per_subap >= 0))
+        self.assertTrue(total_counts[0] >= 0)
+        self.assertTrue(subap_counts[0] >= 0)
```
alfiopuglisi: Should also test the generation time, like for the pyramid

specula/processing_objects/pyr_slopec.py
guidoagapito: you are right, removed lines 84 and 85

specula/processing_objects/pyr_slopec.py
guidoagapito: you are right, removed

specula/processing_objects/sh_slopec.py
guidoagapito: you are right, removed

test/test_sh_slopec.py
guidoagapito: done

FabioRossiArcetri (approved):

FabioRossiArcetri (approved):

PR #417, opened 2025-10-30
(GA) update_interval in Atmo random with tests

guidoagapito:


FabioRossiArcetri (commented):
Why do we need this?

guidoagapito:
We need this to perform a calibration in partial correction. In this case we want to have the same turbulence during push and pull.

FabioRossiArcetri:
> We need this to perform a calibration in partial correction. In this case we want to have the same turbulence during push and pull.

Could it be done in the post trigger? Also, it should be done in the InfinniteAtmosphere as well, just in case we will want to use it in the same context at some point.

guidoagapito:
> > We need this to perform a calibration in partial correction. In this case we want to have the same turbulence during push and pull.
> 
> Could it be done in the post trigger? Also, it should be done in the InfinniteAtmosphere as well, just in case we will want to use it in the same context at some point.

Moved to prepare and post trigger. 

The same approach cannot work in infinite phases screen without some changes, because in infinite phases screen wind speed and time is used to define the new position. So, I would not like to do these changes in this pull request.

FabioRossiArcetri (approved):

PR #419, opened 2025-10-31
Added overwrite to m2c and sn_calibrator

Menessao:


specula/processing_objects/sn_calibrator.py
```diff
@@ -35,7 +35,7 @@ def trigger_code(self):
             self.slopes.slopes += self.local_inputs['in_slopes'].slopes.copy()
         self._n_iter += 1
 
-    def finalize(self):
+    def finalize(self,overwrite:bool=False):
```
guidoagapito, line 38: overwrite should be an argument of the init method and can be passed to finalize with self.
See as an example https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/specula/processing_objects/im_calibrator.py
Could you also add a test like test_existing_im_file_with_overwrite in https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/test/test_im_rec_calibrator.py ?

PR #420, opened 2025-11-01
Overwrite fix

Menessao:
overwrite is now an attribute of the sn_calibrator base processing object. I have also added a test_sn_calibrator.py script, brutally copied from test_im_rec_calibrator.py with appropriate changes. I am ignorant in python and pytests, so let me know if I was supposed to do something different

### DIFF
--- a/test/test_sn_calibrator.py
+++ b/test/test_sn_calibrator.py
@@ -31,7 +31,8 @@ class TestImRecCalibrator(unittest.TestCase):
             f.write('')
 
         with self.assertRaises(FileExistsError):
-            _ = SnCalibrator(data_dir=self.test_dir, output_tag=sn_tag)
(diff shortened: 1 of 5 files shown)


test/test_sn_calibrator.py
```diff
@@ -0,0 +1,47 @@
...
+
+import os
+import tempfile
+import unittest
+import shutil
+
+from specula.processing_objects.sn_calibrator import SnCalibrator
+
+
+class TestImRecCalibrator(unittest.TestCase):
```
guidoagapito: Please change the name of this class.

guidoagapito:
The pull request is fine except a minor point. Thanks!

guidoagapito (approved):

PR #421, opened 2025-11-03
(GA) Spot monitor with test

guidoagapito:


FabioRossiArcetri (approved):
Nice to have this. Considering the fitting is done in cpu with astropy, at some point we will  probably want to implement our own agnostic fitting to support cpu/gpu computation.

PR #423, opened 2025-11-05
allow extra_delta_time for atmo evolution to be different for each layer

bernadett92:
I wanted to set a different extra_delta_time for each layer, which was not possible before.


guidoagapito (approved):

alfiopuglisi (commented):
I like that it works without touching the trigger code!
It would be nice to have some tests actually exercising the new feature: setting up an AtmoEvolution object with a time vector and verifying that self.delta_time elements are increasing correctly (each according to its own time step)

bernadett92:
I agree! I'll add some tests for the new feature.

alfiopuglisi (approved):

PR #428, opened 2025-11-08
Alpao geometry

Menessao:


specula/lib/compute_zonal_ifunc.py
```diff
@@ -18,15 +18,19 @@ def compute_zonal_ifunc(dim, n_act, xp=np, dtype=np.float32, circ_geom=False, an
     step = float(dim) / float(n_act)
 
     # ----------------------------------------------------------
+    if circ_geom is True:
```
guidoagapito, line 21: here we should check if circ_geom and geom are both set and give an error message

guidoagapito:
We need a test file, it would be enough a file with four tests, one for each geom and one for a not existing geom like:

```
def test_invalid_geom_raises():
    with pytest.raises(ValueError):
        compute_zonal_ifunc(dim=32, n_act=4, geom='not_a_geom', xp=np)
```
 
 The standard tests can simply check that the output is valid and the number of actuators is the expected value.

specula/lib/compute_zonal_ifunc.py
```diff
@@ -37,20 +45,28 @@ def compute_zonal_ifunc(dim, n_act, xp=np, dtype=np.float32, circ_geom=False, an
...
+        
         # Convert from polar to Cartesian coordinates
         x = pol_coords[1] * xp.cos(xp.radians(pol_coords[0])) + x_c
         y = pol_coords[1] * xp.sin(xp.radians(pol_coords[0])) + y_c
-
-        # Maximum radius (outer boundary)
-        R = pol_coords[1].max()  # The maximum radial value is the outer boundary
-    else:
+    
+    elif geom == 'alpao':
```
guidoagapito, line 53: Here you need to add n_act_tot or before line 76 you should count n_act_tot from x.
In the second case you should remove the lines where n_act_tot is computed for the different geometries.

test/test_compute_zonal_ifunc.py
```diff
@@ -0,0 +1,40 @@
...
+
+import specula
+specula.init(0)
+
+from specula.lib.compute_zonal_ifunc import compute_zonal_ifunc
+
+
+class TestComputeZonalIfunc(unittest.TestCase):
+
+  def test_invalid_geom_raises(self):
```
alfiopuglisi: These tests are run on CPU only, not on GPU. Is this intentional? See most test files like test_base_operation.py to see how to run them on GPU as well.

alfiopuglisi:
A generic comment for the future: I see that this PR was done from a local branch, while instead one should:
- fork the repo
- send a PR
- merge using "Squash and merge"

specula/lib/compute_zonal_ifunc.py
```diff
@@ -18,16 +18,23 @@ def compute_zonal_ifunc(dim, n_act, xp=np, dtype=np.float32, circ_geom=False, an
...
-            na = xp.arange(round((n_act + 1) / 2)) * 6
+            na = xp.arange(xp.ceil((n_act + 1) / 2)) * 6
         else:
             step *= float(n_act) / float(n_act - 1)
-            na = xp.arange(round(n_act / 2)) * 6
+            na = xp.arange(xp.ceil(n_act / 2)) * 6
         na[0] = 1  # The first value is always 1
-
-        n_act_tot = int(xp.sum(na))
+        n_act_tot = xp.sum(na)
```
guidoagapito, line 37: n_act_tot must be a integer, see line 76:

ifs_cube = xp.zeros((n_act_tot, dim, dim), dtype=dtype)

specula/lib/compute_zonal_ifunc.py
```diff
@@ -37,20 +44,29 @@ def compute_zonal_ifunc(dim, n_act, xp=np, dtype=np.float32, circ_geom=False, an
...
+    
+    elif geom == 'alpao':
+        x, y = xp.meshgrid(xp.linspace(0, dim, n_act), xp.linspace(0, dim, n_act))
+        x, y = x.ravel(), y.ravel()
+        x_c, y_c = dim / 2, dim / 2 # center
+        rho = xp.sqrt((x-x_c)**2+(y-y_c)**2)
+        rho_max = (dim*(9/8-n_act/(24*16)))/2 # slightly larger than dim, depends on n_act
+        x = x[rho<=rho_max]
+        y = y[rho<=rho_max]
+        n_act_tot = xp.size(rho)
```
guidoagapito, line 61: see other comment

guidoagapito:
I suggest running the test before implementing any changes:

`python -m unittest test_compute_zonal_ifunc.py`

PR #429, opened 2025-11-10
Physical propagation

bernadett92:
Fixed bug in physical propagation code and verified it against PASSATA implementation.


specula/processing_objects/atmo_propagation.py
```diff
@@ -157,20 +158,17 @@ def trigger_code(self):
...
                     tmp_phase = interpolator.interpolate(layer.phaseInNm)
-                    if self.doFresnel:
-                        tmp_phase = self.physical_propagation(tmp_phase, li)
+
                     output_ef.A *= interpolator.interpolate(layer.A)
                     output_ef.phaseInNm += tmp_phase
 
+                if self.doFresnel:
+                    tmp_ef = self.physical_propagation(output_ef.ef_at_lambda(self.wavelengthInNm), li)
+                    output_ef.phaseInNm = self.xp.angle(tmp_ef) * self.wavelengthInNm / (2 * self.xp.pi)
```
alfiopuglisi: These two assignments should use this syntax:

output_ef.phaseInNm[:] = ...

in order to avoid re-allocating the phaseInNm array every time (same goes for output_ef.A). This will make the code more efficient on GPUs. In general operations should be done in-place as much as possible.

alfiopuglisi (approved):

PR #437, opened 2025-11-12
(GA) PlotVectorDisplay and crop in pixelsDisplay with test and precision in baseValue

guidoagapito:


test/test_display.py
```diff
@@ -66,7 +68,7 @@ def test_pixels_display_init_and_trigger(self, target_device_idx, xp):
         """Test PixelsDisplay initialization and trigger"""
         pixels_data = xp.arange(9).reshape((3,3))
         pixels = Pixels(3, 3, bits=16, signed=0, target_device_idx=target_device_idx)
-        pixels.set_value(pixels_data)
+        pixels.value=pixels_data
```
alfiopuglisi: Using set_value() is intentional and should not be replaced by direct .value access (in fact this is wrong because it is replacing instead of overwriting).
Same for 5 other occurrences in this PR

guidoagapito: Done, I had to update base_value ini to convert with to_xp value argument

guidoagapito: Other changes:
- BIG ONE - precision argument in baseValue to correctly manage precision option of processing objects that generate the baseValue --> this means updates in several classes
- Correct management of scalar in init and set_value of baseValue

FabioRossiArcetri (approved):

alfiopuglisi (approved):

PR #443, opened 2025-11-17
(GA) special yaml options _data and _object

guidoagapito:


docs/simulation_basics.rst
```diff
@@ -65,6 +65,46 @@ Configuration System
...
+
+.. code-block:: yaml
+
+   pupil_data: "pupil_mask.fits"      # Loads the pupil mask from a FITS file
+   atmo_data: "atmo_layers.fits"      # Loads atmospheric layers
+
+``<name>_object`` option
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Restores a full data object from disk (typically from a FITS file).
```
alfiopuglisi, line 88: Let's add explicitly that: the object class is inferred by the type hint on the relevant initialization parameter as defined in the .py class source file, and the class' restore() method is called using the tag or filename as an argument (maybe ask an AI to rephrase this correctly)

guidoagapito: I added:

`- The object class is automatically inferred from the type hint of the corresponding initialization parameter in the Python class definition. When restoring an object, SPECULA calls the class's ``restore()`` method, passing the specified tag or filename as an argument.`

alfiopuglisi (approved):

PR #450, opened 2025-12-07
(GA) calc_geom bugfix and new tests

guidoagapito:

PR #451, opened 2025-12-07
Coronography

Menessao:
Adding common corongraphs as processing objects to SPECULA

A coronograph abstract class is created, supporting any coronograph type (including the perfect coronograph, if desired)
The other corongraphs implemented are:

- Lyot coronograph (with knife-edge or dot/anular focal plane mask)
- vortex coronograph (with option for inner vortex)
- 4 quadrant coronograph
- phase plate apodization (APP) coronograph 
- phase-induced amplitude apodization pupil Lyot coronograph (PAPLC, an APP followed by a Lyot)


Menessao:
Still a Work in Progress, but I am requesting a review anyway to clear up a few points:

1. FOV. I noticed in the focal plane filter there was a lot of work to make it compatible with a FoV with improved resolution. I could not get my head around it, so I simplified things a bit but lost that functionality. I can try to add it back in if required

2. TESTS. So far I have only implemented tests for the Lyot coronograph, blatantly copying what was done for the focal_plane_filter. If this is ok, I can implement the same for the other coronographs.

3. WAVELENGTH. All coronographs are defined for a specific wavelength, that defines mask size/phase shift. I am not sure on how to best deal with an input electric field given at a different wavelength. Should I access that information directly from the EF object?

guidoagapito:
> Still a Work in Progress, but I am requesting a review anyway to clear up a few points:
> 
> 1. FOV. I noticed in the focal plane filter there was a lot of work to make it compatible with a FoV with improved resolution. I could not get my head around it, so I simplified things a bit but lost that functionality. I can try to add it back in if required

I think that FocalPlaneFilter could become the coronagraph abstract class with a few changes. In this way you get all its functionalities.
What should be missing are:

- Apodize electric field
- Apply pupil stop
- make_focal_plane_mask
- make_pupil_plane_mask
- make_apodizer

I think that "Apodize electric field" and "Apply pupil stop" can be combined in a single step.
Furthermore, they should be applied to self.ef_in before padding.
I proposed these steps for the trigger:
1. Apodize electric field and apply pupil stop
2. Pad the electric field
3. Propagate field to focal plane with FFT
4. Apply focal plane mask (appropriately shifted)
5. Return to the pupil plane with IFFT

So, minimum changes to FocalPlaneFilter should be required to get this.

> 2. TESTS. So far I have only implemented tests for the Lyot coronograph, blatantly copying what was done for the focal_plane_filter. If this is ok, I can implement the same for the other coronographs.

Basic tests are good for now:
- output size
- output amplitude
- flat input phase --> flat output phase
- S0
- PSF (comparison with and without coronagraph?)

> 3. WAVELENGTH. All coronographs are defined for a specific wavelength, that defines mask size/phase shift. I am not sure on how to best deal with an input electric field given at a different wavelength. Should I access that information directly from the EF object?

As for the pyramid and the SH and WFS in general: coronagraphs are monochromatic for a given wavelength. The electric field is unaware of the wavelength. The issue arises if the output of the coronagraph is passed to a PSF (or a different object) that operates at a different wavelength.
We have two options:

1. We add the PSF computation inside the coronagraph class
2. We add a new optional property to the electric field: wavelength. This property by default is None and the behaviour of the  electric field is not changed in this case. Instead the coronagraph change the property from None to the  wavelength and calling `ef.phi_at_lambda(self.wavelengthInNm)` with a different wavelength to the one defined by this object should raise an error.

@alfiopuglisi what do you think?

Menessao:
> I think that FocalPlaneFilter could become the coronagraph abstract class with a few changes. In this way you get all its functionalities. What should be missing are:
> 
> * Apodize electric field
> * Apply pupil stop
> * make_focal_plane_mask
> * make_pupil_plane_mask
> * make_apodizer
> 
> I think that "Apodize electric field" and "Apply pupil stop" can be combined in a single step. Furthermore, they should be applied to self.ef_in before padding. I proposed these steps for the trigger:
> 
> 1. Apodize electric field and apply pupil stop
> 2. Pad the electric field
> 3. Propagate field to focal plane with FFT
> 4. Apply focal plane mask (appropriately shifted)
> 5. Return to the pupil plane with IFFT
> 
> So, minimum changes to FocalPlaneFilter should be required to get this.
> 

Updated as suggested.
However, I left the pupil stop after step 5, as I think the focal plane mask of some coronographs diffracts the light outside the pupil (e.g. vortex), so placing the pupil stop upstream might not work.
If this is not the case, pardon my poor knowledge of optics, and I will merge pupil stop and apodization in a single step before the focal plane mask.

> Basic tests are good for now:
> 
> * output size
> * output amplitude
> * flat input phase --> flat output phase
> * S0
> * PSF (comparison with and without coronagraph?)

Implemented them for vortex, lyot, four quadrant. Still working on the APP

> 
> As for the pyramid and the SH and WFS in general: coronagraphs are monochromatic for a given wavelength. The electric field is unaware of the wavelength. The issue arises if the output of the coronagraph is passed to a PSF (or a different object) that operates at a different wavelength. We have two options:
> 
> 1. We add the PSF computation inside the coronagraph class
> 2. We add a new optional property to the electric field: wavelength. This property by default is None and the behaviour of the  electric field is not changed in this case. Instead the coronagraph change the property from None to the  wavelength and calling `ef.phi_at_lambda(self.wavelengthInNm)` with a different wavelength to the one defined by this object should raise an error.
> 
> @alfiopuglisi what do you think?

Would it be possible to define a bandwidth of operation around the reference wavelength? Do you have something similar for WFSs?


Further questions (easier this time):

1.  Override: do you have a decorator for the overridden methods of inherited classes? I know typings.override gives issues in some Python versions, so this does not work. It is only cosmetic anyway, it's just to mark which functions have been overwritten.
2. Do you want me to adapt the perfect coronograph so that it also is inherited from the new abstract coronograph class?
3. Does it make sense to do something similar for the focal plane filter?

guidoagapito:
Ok for the pupil stop as last step, it is true that it is not equivalent to a pupil stop in the first step.

For the wavelength let's wait Alfio comments.

Answer to further questions:

- We don't currently use an @override decorator in SPECULA. Since it's purely cosmetic/documentary and can cause compatibility issues across Python versions, we've not included it.
- I would not modify the perfect coronagraph for now.
- I would change focal plane filter to inherit the abstract class.

Menessao:
There might still be issues with the FoV resolution: I still do not fully understand it, nor have come up with a neat solution to compute the correct resolution for a given lambda/D angle (e.g. fov_res = 4 for lambda/D = 0.25,1.25,2.25 ... ?)

Other than that, I think everything is working
image
image
image
image
image

guidoagapito:
> There might still be issues with the FoV resolution: I still do not fully understand it, nor have come up with a neat solution to compute the correct resolution for a given lambda/D angle (e.g. fov_res = 4 for lambda/D = 0.25,1.25,2.25 ... ?)

fov_res is not a user parameter — it's calculated automatically by the calc_geometry method to ensure the internal FoV meets the requested FoV.

Here's what happens:

1. Natural FoV
The "natural" FoV is determined by the input pixel pitch:
```
Fov_natural = lambda / pixel_pitch  # in radians
```
This corresponds to the maximum spatial frequency present in the electric field. According to the Nyquist theorem, the pixel pitch defines the sampling limit of the wavefront.

2. fov_res (interpolation factor)
Purpose: Interpolates the input phase array if the natural FoV is too small for the requested FoV
Calculated as:
```
Fov_internal = lambda / D  # natural FoV based on aperture diameter
minfov = FoV * (1 - fov_errinf)  # minimum acceptable FoV

if Fov_internal < minfov:
    fov_res = ceil(minfov / Fov_internal)  # interpolation factor
else:
    fov_res = 1  # no interpolation needed
```
Effect: The input phase array is interpolated by fov_res, increasing its size from DpupPix to DpupPix * fov_res
Important: The interpolation does not provide extra spatial frequencies — it only upsamples the existing data. Its purpose is to avoid numerical issues when applying a focal plane mask that would be larger than the focal plane array without interpolation.
Example: If natural FoV = 1.0 arcsec but you request 3.0 arcsec with fov_errinf=0.5 → minfov = 1.5 → fov_res = 2

fft_res (default = 3.0) is the parameter that controls the PSF sampling resolution in the focal plane: `lambda / (D * fft_res)` per pixel. This is the parameter you should adjust if you want to change the focal plane resolution.

The relationship is:

fov_res → adjusts the field size (FoV) via input interpolation
fft_res → adjusts the PSF sampling (resolution) via FFT oversampling

specula/processing_objects/abstract_coronograph.py
```diff
@@ -0,0 +1,285 @@
...
+from abc import abstractmethod
+
+class Coronograph(BaseProcessingObj):
+    def __init__(self,
+                 simul_params: SimulParams,
+                 wavelengthInNm: float,
+                 fov: float,
+                 fov_errinf: float = 0.1,
+                 fov_errsup: float = 10,
+                 fft_res: float = 4.0,
```
guidoagapito: the default value in PWFS and focal plane filter was 3, is there a reason to change it to 4?

specula/processing_objects/abstract_coronograph.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+    def calc_geometry(self, DpupPix, pixel_pitch, lambda_, FoV,
+                      fov_errinf=0.1, fov_errsup=0.5,  fft_res=3.0):
+        D = DpupPix * pixel_pitch
+        Fov_internal = lambda_ * 1e-9 / D * (D / pixel_pitch) * RAD2ASEC
+        minfov = FoV * (1 - fov_errinf)
+        maxfov = FoV * (1 + fov_errsup)
+        fov_res = 1.0
+        if Fov_internal < minfov:
```
guidoagapito: in #450 focal_plane_filter.py calc_geometry was update to fix some issue seen in the modulated pyramid. Please follow the same approach.

guidoagapito:
to be merged after #450

specula/processing_objects/abstract_coronograph.py
```diff
@@ -0,0 +1,318 @@
...
+        fov_errinf=0.1,
+        fov_errsup=0.5,
+        fft_res=3.0):
+
+        fov_internal = lambda_ * 1e-9 / pixel_pitch * RAD2ASEC
+
+        minfov = FoV * (1 - fov_errinf)
+        maxfov = FoV * (1 + fov_errsup)
+
+        fov_res = 1.0
```
guidoagapito: following Alfio's comments some changes have been made to #450 that should be "imported" here, thanks.

guidoagapito:
There are some conflicts with the focal_plane_filter. Merging main into this branch would solve these conflicts.

guidoagapito (approved):

specula/processing_objects/abstract_coronograph.py
```diff
@@ -0,0 +1,314 @@
...
+        ef_pad[pad_start:pad_start+self.fft_sampling, 
+                    pad_start:pad_start+self.fft_sampling] = pup_ef
+        return self.xp.fft.fft2(ef_pad)
+    
+    # def _focal_to_pupil_plane(self, fp_ef):
+    #     ef_pad = self.xp.fft.ifft2(fp_ef)
+    #     pad_start = self.fft_padding // 2
+    #     ef_pp = ef_pad[pad_start:pad_start+self.fft_sampling, 
+    #                 pad_start:pad_start+self.fft_sampling]
+    #     return ef_pp
```
alfiopuglisi: Is it necessary to keep this function commented out? If they are useful, it can be uncommented

specula/processing_objects/abstract_coronograph.py
```diff
@@ -0,0 +1,314 @@
...
+                precision=self.precision
+            )
+
+            # Create the interpolator (like in SH)
+            self.interp = Interp2D(
+                in_ef.size,
+                (self.fft_sampling, self.fft_sampling),
+                0, #-self.rotAnglePhInDeg,  # Negative angle for PASSATA compatibility
+                0, #self.xShiftPhInPixel,
+                0, #self.yShiftPhInPixel,
```
alfiopuglisi: Instead of using comments, these can be keyword arguments (rotInDeg=0, etc)

specula/processing_objects/apodizer_coronograph.py
```diff
@@ -0,0 +1,228 @@
...
+                         target_device_idx=target_device_idx, 
+                         precision=precision)
+        apodizer_phase, self.target_contrast = self.define_apodizing_phase(pupil, contrastInDarkHole,
+                                                          iwaInLambdaOverD, owaInLambdaOverD, beta,
+                                                          symmetric_dark_hole=make_symmetric, max_its=max_its)
+        self.apodizer = self.xp.exp(1j*apodizer_phase, dtype=self.xp.complex64)
+        
+
+    def define_apodizing_phase(self, pupil, contrast,
+                               iwa, owa, beta:float,
```
alfiopuglisi: Are iwa and owa float as well?

specula/processing_objects/apodizer_coronograph.py
```diff
@@ -0,0 +1,228 @@
...
+                            symmetric_dark_hole:bool=False, 
+                            max_its:int=1000):
+        target_contrast = self.xp.ones([self.fft_totsize,self.fft_totsize],dtype=float)
```
alfiopuglisi: dtype for self.xp.ones() should be "self.dtype", so that single and double precision is handled automatically

specula/processing_objects/apodizer_coronograph.py
```diff
@@ -0,0 +1,228 @@
...
+        pad_pupil = self.xp.zeros([self.fft_totsize, self.fft_totsize])
+        pad_pupil[pad_start:pad_start+self.fft_sampling, 
+                    pad_start:pad_start+self.fft_sampling] = self.xp.array(pupil)
+        app = generate_app_keller(pad_pupil, self.xp.array(target_contrast), max_iterations=max_its, beta=beta, xp=self.xp)
+        apodizer_phase = self.xp.zeros(pupil.shape)
+        apodizer_phase[pupil>0] = self.xp.angle(app)[pad_pupil>0.0]
+        return apodizer_phase, target_contrast
+
+    def make_focal_plane_mask(self):
+        return self.xp.ones([self.fft_totsize,self.fft_totsize])
```
alfiopuglisi: Plese add "dtype=self.dtype". This goes for all xp.ones() and xp.zeros() (there are many)

specula/processing_objects/lyot_coronograph.py
```diff
@@ -0,0 +1,69 @@
...
+                             f' outer size is {outerStopAsRatioOfPupil*1e+2:1.0f}% of pupil')
+        
+        if knife_edge is True and owaInLambdaOverD is not None:
+            raise ValueError('OWA cannot be defined for the knife-edge focal plane mask')
+        
+        fov = wavelengthInNm * 1e-9 / simul_params.pixel_pitch * RAD2ASEC
+        # if iwaInLambdaOverD > 0.0:
+        #     fov *= max(1.0,1/(iwaInLambdaOverD-(iwaInLambdaOverD % 1)))
+
+        # fft_res = max(1/(iwaInLambdaOverD-np.mod(iwaInLambdaOverD,1)), 1/(owaInLambdaOverD-np.mod(owaInLambdaOverD,1)))
```
alfiopuglisi: If fft_res can be automatically calculated, then its default value could be "None" and, if None, be calculated as in the commented-out code. If this feature instead is not desired, then we should remove these comments.

specula/processing_objects/vortex_coronograph.py
```diff
@@ -0,0 +1,75 @@
...
+        X, Y = self.xp.mgrid[0:N, 0:N]   
+        theta = self.xp.arctan2((X - c), (Y - c))
+        theta = (theta + 2 * self.xp.pi) % (2 * self.xp.pi)
+        vortex = self._charge * theta
+        if self._inVortex is True:
+            rho = self.xp.sqrt((X-c)**2+(Y-c)**2)
+            inVortex = self._innerCharge * theta + self._innerShift
+            inRho = self._innerRadInLambdaOverD * self.fft_res
+            vortex[rho<=inRho] = inVortex[rho<=inRho]
+        fp_mask = self.xp.exp(1j*vortex, dtype=self.xp.complex64)
```
alfiopuglisi: dtype should be "self.complex_dtype". There are three self.xp.exp() calls in this PR where this comment applies

alfiopuglisi (changes requested):
Please see the inline comments

alfiopuglisi (approved):

PR #452, opened 2025-12-08
(GA) bugfixes in extended object

guidoagapito:
to be merged after #450


alfiopuglisi (changes requested):
The changes in modulated_pyramid.py are the same of PR#450, already merged, and it might result in a conflict. It's better to undo them before merging.

guidoagapito:
> The changes in modulated_pyramid.py are the same of PR#450, already merged, and it might result in a conflict. It's better to undo them before merging.

I merged the branch with main

alfiopuglisi (approved):

PR #460, opened 2025-12-17
calc_geometry and interpolation factorization

alfiopuglisi:
Factorized out calc_geometry() and EF interpolation calculations into their own functions and classes, which are then used by multiple WFSs.


specula/processing_objects/modulated_pyramid.py
```diff
@@ -601,37 +504,16 @@ def setup(self):
...
-            )
-        else:
-            self._do_interpolation = False
-            # Use the original field directly (like SH does)
-            self._wf_interpolated = in_ef
+        self.ef_interpolator = EFInterpolator(
+            in_ef = in_ef,
+            out_shape=(self.fft_sampling, self.fft_sampling),
+            rotAnglePhInDeg=self.rotAnglePhInDeg,
+            xShiftPhInPixel=self.yShiftPhInPixel,
```
guidoagapito: x and y are swapped

alfiopuglisi: fixed

guidoagapito (approved):

Menessao (approved):
Standardizing and simplifying calc_geom and EF interpolation is great!

My only comment does not have to do with the pull request but rather with the fact that the _toccd_ function might be underused because its name is a bit misleading, but I guess this has always been this way and changing it would be more effort than return.

alfiopuglisi:
> My only comment does not have to do with the pull request but rather with the fact that the _toccd_ function might be underused because its name is a bit misleading, but I guess this has always been this way and changing it would be more effort than return.

Ok will be done in another PR in order not to clutter too much this one

PR #462, opened 2025-12-18
(GA) update to center and step

guidoagapito:


specula/lib/compute_zonal_ifunc.py
```diff
@@ -32,12 +30,17 @@ def compute_zonal_ifunc(dim, n_act, xp=np, dtype=np.float32,circ_geom:bool=False
...
     if geom == 'circular':
         if n_act % 2 == 0:
-            na = xp.arange(xp.ceil((n_act + 1) / 2)) * 6
+            n_act_radius = int(xp.ceil((n_act + 1) / 2))
+            na = xp.arange(n_act_radius) * 6
         else:
-            step *= float(n_act) / float(n_act - 1)
-            na = xp.arange(xp.ceil(n_act / 2)) * 6
+            n_act_radius = int(xp.ceil(n_act / 2))
+            na = xp.arange(n_act_radius) * 6
```
alfiopuglisi: "na" calculation can be moved outside the "if" statement since it is identical

guidoagapito: Done

alfiopuglisi (approved):

Menessao (commented):
Actuator grid now seems centered on the mask: for the ALPAO case, I get better looking KL modes (even with the central obscuration, the last modes have square symmetry). 

I still have the same issues with the circular geometry: for a large number of actuators (Npix=160, n_act=40/41) and oversampling = 2 for the KL computation, the reconstructed KL from the first loop residual display spurious oscillations at high order modes.
image
My machine is not powerful enough to compute the KL modes for Npix=160 with an oversampling factor of 4, which is something that might solve the issue.

One further test that could be implemented would be to check that the number of slaved actuators with a circular pupil and a given central obscuration is the expected one. Simulating a mask with 69 pixels across and a 30% obscuration, slaving with a 0.7 threshold an ALPAO with 9 actuators on the diameter should give 5 slave actuators:
image
 Moving the actuators' y_coordinates by 1 pixels gives an additional slave as result:
image
I could engineer similar test cases for the square and circular geometries, if useful.

guidoagapito:
> Actuator grid now seems centered on the mask: for the ALPAO case, I get better looking KL modes (even with the central obscuration, the last modes have square symmetry).
> 
> I still have the same issues with the circular geometry: for a large number of actuators (Npix=160, n_act=40/41) and oversampling = 2 for the KL computation, the reconstructed KL from the first loop residual display spurious oscillations at high order modes. image My machine is not powerful enough to compute the KL modes for Npix=160 with an oversampling factor of 4, which is something that might solve the issue.
> 
> One further test that could be implemented would be to check that the number of slaved actuators with a circular pupil and a given central obscuration is the expected one. Simulating a mask with 69 pixels across and a 30% obscuration, slaving with a 0.7 threshold an ALPAO with 9 actuators on the diameter should give 5 slave actuators: image Moving the actuators' y_coordinates by 1 pixels gives an additional slave as result: image I could engineer similar test cases for the square and circular geometries, if useful.

Ok, I will merge this pull request and do any additional bugfix in a new PR.

PR #440, opened 2025-11-16
(GA) extended source pyramid with reduced memory footprint

guidoagapito:

PR #466, opened 2026-01-13
(GA) extended pyramid update

guidoagapito:


specula/processing_objects/ext_source_pyramid.py
```diff
@@ -129,48 +143,74 @@ def prepare_trigger(self, t):
...
-        # Reset output arrays for this frame
-        self.pyr_image *= 0
-        self.fpsf *= 0
+        # Get extended source coefficients for current frame (only valid points)
+        coeff_ttf = self.ext_source_coeff.value[self.valid_idx, :3]
+        ffv_valid = self.flux_factor_vector[self.valid_idx]
+        n_valid = self.valid_idx.shape[0]
+
+        chunk_idx = 0
+        while chunk_idx * self.max_batch_size < n_valid:
```
FabioRossiArcetri: We should find a way to avoid this while loop, this would probably result in more roboust stream computation

guidoagapito: This class was introduced because the standard pyramid allocates all the points at once: this means saturating the GPU (or CPU) memory for large extended sources.

You can see the old pull request here:
https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/440

Therefore I think it is unavoidable to split the computation in N sequential parts with a while or a for cycle.

FabioRossiArcetri (approved):

Issue #474, opened 2026-01-28
Missing initialization handling


guidoagapito:
Description
When `specula.init()` is not called before creating SPECULA objects, the global variables `global_precision` and `default_target_device_idx` remain None.
This causes issues when initializing `BaseTimeObj` and any objects that inherit from it.

Current Behavior
If `specula.init()` is not called, objects can be created but may behave unexpectedly or fail with confusing errors.
The uninitialized state is not caught early, leading to difficult debugging.

Proposed Solution
A warning is issued when `BaseTimeObj` detects uninitialized globals, and it falls back to safe defaults (CPU device with single precision).

Alternative Approaches
Raise an error (stricter): Force users to explicitly call `specula.init()` before creating any objects.

@alfiopuglisi and @FabioRossiArcetri:
Should we maintain the current warning approach or switch to raising a RuntimeError to enforce explicit initialization?

guidoagapito:
This issue causes problems when running some tests, the ones without specula.init().
This happens only when tests run individually. 
For this specific point I opened https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/475.

alfiopuglisi:
I propose two possibilities:

1) force the user to call init(). Similar to now, but raise an exception if specula is used without having called init()
2) if the user does not call init(), perform it automatically with sensible defaults (like device=0 and precision=float32), with fallbacks like CPU, etc.

guidoagapito:
The first option seems clearer to me and requires the user to choose a specific configuration.

PR #475, opened 2026-01-29
(GA) specula.init in all tests

guidoagapito:


guidoagapito:
I add specula.init(0) in all the tests where it was not present. This avoid issue with global precision and global device idx. See https://github.com/ArcetriAdaptiveOptics/SPECULA/issues/474.

FabioRossiArcetri (approved):

PR #479, opened 2026-01-31
(GA) new CUDA function for interpolation: interp2_kernel_onthefly

guidoagapito:
Updated GPU interpolation with CUDA: coordinates are not stored to save GPU memory.

I tested it on gandalf.


FabioRossiArcetri (approved):

specula/lib/interp2d.py
```diff
@@ -166,24 +267,52 @@ def interpolate(self, value, out=None):
...
-            if self.output_shape[0] % block[0]:
-                num_blocks_2d += 1
-            grid = (num_blocks_2d, num_blocks_2d)
-
-            if self.dtype == cp.float32:
-                self.interp2_kernel_float(grid, block, (value, out, self.output_shape[1],  self.output_shape[0],  self.input_shape[1], self.input_shape[0], self.xx, self.yy))
-            elif self.dtype == cp.float64:
-                self.interp2_kernel_double(grid, block, (value, out, self.output_shape[1],  self.output_shape[0],  self.input_shape[1], self.input_shape[0], self.xx, self.yy))
+        if self.xp == cp: # pragma: no cover
+            block = (16, 16, 1)
```
alfiopuglisi: I think a 2d grid is enough here, a dimension of length 1 is redundant and in fact the z dimension is not used in the kernel source

specula/lib/interp2d.py
```diff
@@ -166,24 +267,52 @@ def interpolate(self, value, out=None):
...
+            grid = (grid_x, grid_y, 1)
+
+            if not self.use_precomputed:
+                # Usa il kernel ottimizzato per memoria
+                if self.dtype == cp.float32:
+                    self.interp2_kernel_onthefly_float(grid, block, (
+                        value, out,
+                        self.output_shape[1], self.output_shape[0],
+                        self.input_shape[1], self.input_shape[0],
+                        cp.float32(self.scale_x), cp.float32(self.scale_y),
```
alfiopuglisi: Do we need all these castings to cp.float32() and cp.float64()? The call to the old kernel, which is still present here after the else, does not use them at all.

specula/lib/interp2d.py
```diff
@@ -166,24 +267,52 @@ def interpolate(self, value, out=None):
...
-            if self.dtype == cp.float32:
-                self.interp2_kernel_float(grid, block, (value, out, self.output_shape[1],  self.output_shape[0],  self.input_shape[1], self.input_shape[0], self.xx, self.yy))
-            elif self.dtype == cp.float64:
-                self.interp2_kernel_double(grid, block, (value, out, self.output_shape[1],  self.output_shape[0],  self.input_shape[1], self.input_shape[0], self.xx, self.yy))
+        if self.xp == cp: # pragma: no cover
+            block = (16, 16, 1)
+            # Calculate grid size for non-square arrays correctly
+            grid_x = (self.output_shape[1] + block[0] - 1) // block[0]
+            grid_y = (self.output_shape[0] + block[1] - 1) // block[1]
+            grid = (grid_x, grid_y, 1)
```
alfiopuglisi: Same comment, the z axis can be removed

specula/lib/interp2d.py
```diff
@@ -42,10 +43,84 @@ class Interp2D():
...
+                    ycoord += shift_y;
+                    
+                    // Clamp to limits (as in __init__)
+                    if (xcoord < 0) xcoord = 0;
+                    if (ycoord < 0) ycoord = 0;
+                    if (xcoord > in_dx - 1) xcoord = in_dx - 1;
+                    if (ycoord > in_dy - 1) ycoord = in_dy - 1;
+                    
+                    // Bilinear interpolation - SAME LOGIC as interp2_kernel_TYPE
+                    int xin = floor(xcoord);
```
alfiopuglisi: Since the code in interp2_kernel_TYPE is now mostly duplicated, it would be better to move it into a separate function (which can be marked __device__ instead of __global__ since it is only called from the GPU) and call it from both kernels. interp2d_kernel_TYPE becomes just a wrapper to this function extracting x/ycoord from xx/yy, while interp2d_kernel_onthefly_TYPE adds the x/ycoord generation code.

specula/lib/interp2d.py
```diff
@@ -166,24 +267,52 @@ def interpolate(self, value, out=None):
...
-                self.interp2_kernel_double(grid, block, (value, out, self.output_shape[1],  self.output_shape[0],  self.input_shape[1], self.input_shape[0], self.xx, self.yy))
+        if self.xp == cp: # pragma: no cover
+            block = (16, 16, 1)
+            # Calculate grid size for non-square arrays correctly
+            grid_x = (self.output_shape[1] + block[0] - 1) // block[0]
+            grid_y = (self.output_shape[0] + block[1] - 1) // block[1]
+            grid = (grid_x, grid_y, 1)
+
+            if not self.use_precomputed:
+                # Usa il kernel ottimizzato per memoria
```
alfiopuglisi: Translate comment to English

specula/lib/interp2d.py
guidoagapito: Done

specula/lib/interp2d.py
guidoagapito: cast moved to init

specula/lib/interp2d.py
guidoagapito: Done

specula/lib/interp2d.py
guidoagapito: Done

specula/lib/interp2d.py
guidoagapito: Done

specula/lib/interp2d.py
```diff
@@ -1,51 +1,131 @@
...
+                    g_out[y * out_dx + x] = bilinear_interp(g_in, in_dx, in_dy, xcoord, ycoord);
+                }
+            }
+            '''
+        interp2_kernel_float = \
+            cp.RawKernel(interp2_kernel.replace('TYPE', 'float'), name='interp2_kernel_float')
+        interp2_kernel_double = \
+            cp.RawKernel(interp2_kernel.replace('TYPE', 'double'), name='interp2_kernel_double')
+        interp2_kernel_onthefly = r'''
+            // Device function for bilinear interpolation
```
alfiopuglisi: If I understand correctly, there are two copies of this function. Is it possible to have a single copy, maybe in a separate string? That was the goal of making a separate function

guidoagapito: You are right! I updated this part.

alfiopuglisi (approved):

PR #480, opened 2026-02-03
(GA) updates to modalrec and filter classes for pseudo open loop control

guidoagapito:


FabioRossiArcetri (approved):

specula/processing_objects/base_filter.py
```diff
@@ -0,0 +1,134 @@
...
+            precision=precision)
+
+        # Inputs
+        self.inputs['delta_comm'] = InputValue(type=BaseValue)
+        self.inputs['gain_mod'] = InputValue(type=BaseValue, optional=True)
+
+        # Outputs
+        self.outputs['out_comm'] = self.out_comm
+        # This output provides the non-delayed command for POLC use
+        # Please don't use it for drive DM / actuator directly
```
alfiopuglisi, line 59: Why not?
I don't think a processing object has the "right" to say what an output will be used for

alfiopuglisi (commented):
Why not?
I don't think a processing object has the "right" to say what an output will be used for

specula/processing_objects/base_filter.py
guidoagapito: This comment specifies that the second output is the non delayed version of the first one and should not be used to drive the DM.
This can be modified as:

This output provides the non-delayed command.
POLC requires the non-delayed command.
DMs and actuators are usually driven with the delayed output.

alfiopuglisi: ok the new version is better

PR #482, opened 2026-02-03
(GA) zeros_common in EFInterpolator with tests

guidoagapito:


FabioRossiArcetri (approved):

guidoagapito:
@FabioRossiArcetri @alfiopuglisi

Add target_rank support for proper cache management across MPI ranks

- Added target_rank attribute to BaseTimeObj (default: 0)
- Updated Simul.build_objects() to set target_rank after object initialization
- Modified cache keys in SH and EFInterpolator to include target_rank
- This ensures cache arrays are properly separated between different MPI ranks
- Cache is now keyed by (target_device_idx, target_rank, shape, dtype, id(xp))

Previously, cache was shared incorrectly between objects on different MPI ranks
with the same geometrical parameters, potentially causing data contamination.
Now each rank maintains separate cached arrays.

I WAS WRONG!

specula/processing_objects/sh.py
```diff
@@ -85,6 +105,9 @@ def __init__(self,
         self._trigger_geometry_calculated = False
         self._mask_threshold = 1e-3  # threshold to consider a pixel inside the mask
 
+        self.psf = None
+        self.psf_shifted = None
```
alfiopuglisi, line 101: If we decide to initialize all members to None, then we should add here self.ef_row

specula/processing_objects/sh.py
```diff
@@ -28,16 +28,36 @@ class SH(BaseProcessingObj):
 
     __zeros_cache = {}
 
-    def _zeros_common(self, *args, **kwargs):
-        '''
-        Wrapper around self.xp.zeros to enable the reuse cache.
+    @classmethod
+    def _zeros_common(cls, shape, dtype, xp,
+                      target_device_idx, target_rank):
```
alfiopuglisi: I am not convinced that all these parameters are needed:
- target_rank is used with multiple processes: within a single Python instance it will be constant, and there should not be a need of tracking it for the cache.
- xp and target_device_idx go together: if the target is -1, then xp will be np, otherwise cp. We should keep target_device_idx to allocate on different GPUs but the xp module seems redundant
- why a classmethod?

specula/lib/extrapolation_2d.py
```diff
@@ -182,6 +185,42 @@ class EFInterpolator():
     '''
     Interpolate the amplitude and phase of an ElectricField object using edge extrapolation.
     '''
+
+    __ef_cache = {}  # Shared cache for ElectricField objects
+    __zeros_cache = {}  # Shared cache for all EFInterpolator instances
+
+    @classmethod
+    def _zeros_common(cls, shape, dtype, xp,
+                      target_device_idx, target_rank):
```
alfiopuglisi: See the comments _zeros_common in the sh class later on.

test/test_extrapolation_2d.py
```diff
@@ -235,3 +241,276 @@ def test_extrapolation_does_not_fail_with_many_edge_pixels(self):
...
+
+        # Check that the same array is used
+        phase_extrap_id_2 = id(interp2.phase_extrapolated)
+        self.assertEqual(phase_extrap_id_1, phase_extrap_id_2,
+                        "Both interpolators should share the same cached array")
+
+    @cpu_and_gpu
+    def test_cache_different_target_ranks(self, target_device_idx, xp):
+        """
+        Test that cache creates separate entries for different target_ranks.
```
alfiopuglisi: If I understand MPI correctly, different target ranks will *not* happen in a single Python process, so there is no need to test for this. @FabioRossiArcetri can you confirm?

test/test_sh.py
```diff
@@ -109,6 +109,9 @@ def test_zeros_cache(self, target_device_idx, xp):
         pixel_pitch = 0.05
         sh_npix = 6
 
+        # clear cache before test
+        SH._SH__zeros_cache.clear()
```
alfiopuglisi, line 113: This "_ SH __ zeros_cache" member is not present in the main branch, nor in this PR, so I don't think it will work. BTW the "_ SH __" at the beginning in redundant.

specula/processing_objects/sh.py
guidoagapito: Ok, I will remove `target_rank`, `xp` is required because this is a `@classmethod` (but is not required in `key`) and the `__zeros_cache` cache is a class attribute (defined outside `__init__`), so it is shared among all `SH` instances. To access it correctly, I need a `@classmethod`.

alfiopuglisi: There should be no need of using a @classmethod to access a class attribute. Let's talk about this on Monday

test/test_sh.py
guidoagapito: This is the "name mangling" of Python:

__attribute → _ClassName__attribute

So, this allows to access to __zeros_cache in the SH class.

test/test_extrapolation_2d.py
guidoagapito: You are correct, MPI process has separate memory, I forgot this point!

specula/processing_objects/sh.py
guidoagapito: Done

guidoagapito:
Removed target_rank management.

specula/processing_objects/sh.py
guidoagapito: Ok, you are right, we can have a classmethod, a staticmethod or a "standard" method.
They should be almost equivalent, but I think @classmethod help to clarify that this is a class-level operation.
However I can modify it if you prefer.

FabioRossiArcetri (approved):

alfiopuglisi (approved):

PR #487, opened 2026-02-06
PSF verbose + ARTE radial_profile

Menessao:
Two minor updates: verbose option for PSF (to avoid printing SR more than one time per iteration if multiple PSF/coro PSFs are computed) and radial_profile from ARTE


guidoagapito:
I disagree with adding a large number of configuration files to the specula repository.
I consider the config part to be just a section with examples, not an exhaustive set of scripts and parameters files.

I think the complete set of files of RISTRETTO can be stored on GitHub in a different repository.

Could you clarify what the purpose of this pull request is? I mean: which of the two cases we are dealing with?

@FabioRossiArcetri @alfiopuglisi Do you have a different opinion on this matter?

Menessao:
> I disagree with adding a large number of configuration files to the specula repository. I consider the config part to be just a section with examples, not an exhaustive set of scripts and parameters files.
> 
> I think the complete set of files of RISTRETTO can be stored on GitHub in a different repository.
> 
> Could you clarify what the purpose of this pull request is? I mean: which of the two cases we are dealing with?
> 

I guess the outcome of this pull request was for me to learn that I should store those files in a different repository. 

If you think the other two changes (verbose boolean option on PSF and coro_PSF and radial_profile from ARTE) are worth a PR I can adjust this so we can merge only those, otherwise I simply scratch everything and apologise.

guidoagapito:
I think the verbose boolean option on PSF and coro_PSF and radial_profile are good.

specula/lib/radial_profile.py
```diff
@@ -0,0 +1,40 @@
+import numpy as np
```
guidoagapito, line 1: I suggest to use xp and dtype arguments as in https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/main/specula/lib/make_xy.py and other library function.
np and np.float64 can be the default values for these arguments.

specula/_version.py
```diff
@@ -0,0 +1,34 @@
+# file generated by setuptools-scm
```
guidoagapito: Is this file required?

alfiopuglisi:
I am late to the party here :) But I noticed that no one is using the radial_profile function, and there is no test for it. Is someone planning to use it soon?

Menessao:
I have thought it would be useful to compute the radial profile of the PSF
(e.g. to evaluate the performance of a coronagraph)

But if there is a better way to do this or nobody besides me is using the
function then I guess it can be pruned.

PR #490, opened 2026-02-09
(GA) error for missing simul_params in LaserLaunchTelescope

guidoagapito:


specula/data_objects/laser_launch_telescope.py
```diff
@@ -92,6 +96,7 @@ def from_header(hdr, target_device_idx=None):
             raise ValueError(f'Error: unknown version {version} in header')
 
         llt = LaserLaunchTelescope(
+            simul_params=SimulParams(zenithAngleInDeg=0.0),
```
alfiopuglisi, line 99: This simul_params is not used

guidoagapito: if `len(tel_position) != 0` (`tel_position` here is `[hdr['TELPOS_X'], hdr['TELPOS_Y'], hdr['TELPOS_Z']]`) we get a value error.
This value error is required to avoid to set an LaserLaunchTelescope object without zenithAngleInDeg when it models a standard laser launch telescope and it is not simply used to get a gaussian kernel for the SH.

alfiopuglisi: I made a mistake, I thought that simul_params was an independent variable instead of parameter to LaserLaunchTelescope()

PR #491, opened 2026-02-09
Physical propagation update

bernadett92:
Fixed bug in zero padding and improved bandlimit filter computation


guidoagapito (approved):

alfiopuglisi:
Good for the bugfix, but I noticed that there are no tests for this feature (or the existing ones do not test enough, otherwise checks wouldn't have passed). They should be added in order to be sure that future changes don't inadvertently introduce some other bugs.

bernadett92:
I agree! I made the tests more sensitive here https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/496#issue-3957128264

PR #492, opened 2026-02-12
(GA) kernels precision and free memory in prepare_for_sh

guidoagapito:


specula/data_objects/convolution_kernel.py
```diff
@@ -352,6 +352,9 @@ def prepare_for_sh(self, sodium_altitude=None, sodium_intensity=None, current_ti
                 self.save(full_path)
                 print('Done')
 
+            # free memory
+            self.real_kernels = None
+
```
alfiopuglisi, line 367: I must admit that I don't understand how this works: kernels are deallocated here, and are only allocated again in restore() or set_value(). But who is calling those methods?

guidoagapito: For the kernel, we allocate two arrays, one real and one complex.
The real array is used for initialization, saving, and restoring, while the complex array is used in the SH.
Therefore, when we prepare the kernel for the SH, the real array becomes superfluous and can be eliminated.

restore is called in prepare_for_sh method, set_value is currently not used, but it was defined to be coherent with other data_objects.

alfiopuglisi: Ok, then in this PR, the save() method must also check if real_kernels is valid, like in get_value()

specula/processing_objects/sh.py
```diff
@@ -284,7 +284,8 @@ def _calc_trigger_geometry(self, in_ef):
                                                             return_fft = True,
```
FabioRossiArcetri, line 284: shall we take the chance to rename this function to _trigger_calc_geometry?

guidoagapito: renamed to _calc_geometry

FabioRossiArcetri (approved):

alfiopuglisi (approved):

PR #502, opened 2026-02-21
(GA) SprintPyr with tests

guidoagapito:


specula/processing_objects/base_sprint_estimator.py
```diff
@@ -238,7 +263,8 @@ def setup(self):
         # Extract DM parameters
         self.ifunc_3d = cpuArray(self.dm.ifunc_obj.ifunc_2d_to_3d(normalize=True))
         self.ifunc_3d = self.ifunc_3d[:, :, self.modes_index]  # Extract only requested modes
-        self.pup_mask = cpuArray(self.dm.mask)
+        if self.pupil_mask is None:
+            self.pup_mask = cpuArray(self.dm.mask)
```
alfiopuglisi: Can we find different names for "self.pupil_mask" and "self.pup_mask" ? They are very similar and could be confused

specula/processing_objects/sprint_pyr.py
```diff
@@ -0,0 +1,377 @@
...
+
+            sl.single_mask = full_mask
+            sl.display_map = idx0_valid
+
+        # get2d() handles the reshaping based on the lengths of mask/display_map.
+        # It returns a single 2D array if slopes_from_intensity=True,
+        # or a list of two 2D arrays [frame_x, frame_y] for standard slopes.
+        frames2d = sl.get2d()
+
+        # Normalizziamo l'output su CPU per matplotlib
```
alfiopuglisi: Translate to English

specula/processing_objects/base_sprint_estimator.py
guidoagapito: we can use pupil_mask for both, no need to have two different names. Changed.

specula/processing_objects/sprint_pyr.py
guidoagapito: Done.

FabioRossiArcetri (approved):

alfiopuglisi (approved):

PR #507, opened 2026-02-25
(GA) reduce memory usage in Zernike generation

guidoagapito:


FabioRossiArcetri (approved):

specula/lib/zernike_generator.py
```diff
@@ -170,7 +170,12 @@ def _rnm_jacobi(self, radialDegree, azimuthalFrequency, rhoArray):
             rho_cpu = np.where(rho_cpu < 0, 0, rho_cpu)
             K = (n - abs(m)) // 2
             const = 1 / eval_jacobi(K, 0, abs(m), 1.0)
-            cJ = eval_jacobi(K, 0, abs(m), 2 * rho_cpu**2 - 1)
+            # in place
```
alfiopuglisi: This in-place calculation looks like overkill wrt. the possible benefits

alfiopuglisi (changes requested):

specula/lib/zernike_generator.py
guidoagapito: Done

alfiopuglisi (approved):

PR #508, opened 2026-03-01
(GA) infinite phasescreen backups and factorization

guidoagapito:


FabioRossiArcetri (approved):

FabioRossiArcetri (approved):

alfiopuglisi:
It looks to me that this PR will use more memory (because of the pre-allocation of all screens) and there is also an extra copy (in line 300,  self.xp.copyto() was previously a reference copy). What is the advantage?

guidoagapito:
> It looks to me that this PR will use more memory (because of the pre-allocation of all screens) and there is also an extra copy (in line 300, self.xp.copyto() was previously a reference copy). What is the advantage?

You are right! I removed the copies, now it uses only references.

alfiopuglisi (approved):

PR #510, opened 2026-03-02
(GA) BaseInserter with tests

guidoagapito:


alfiopuglisi:
A general point: we have other processing objects (like the DM and Modalrec) where some parameters are used for slicing or inserting the inputs or outputs. We also have a BaseSlicer that does not seem to be used. We could think of a single mechanism for all those

alfiopuglisi (approved):

alfiopuglisi:
> A general point: we have other processing objects (like the DM and Modalrec) where some parameters are used for slicing or inserting the inputs or outputs. We also have a BaseSlicer that does not seem to be used. We could think of a single mechanism for all those

For example the modal_rec.py line 102 is very similar to what the BaseSlicer is doing to compute the indexes

specula/processing_objects/base_inserter.py
```diff
@@ -0,0 +1,66 @@
...
+    ----------
+    output_size : int
+        Size of the large output vector.
+    indices : list of [src_indices, dest_indices] pairs, optional
+        Each pair defines explicit indices in the input and output vectors.
+        Example: indices=[[0,1,2], [1,3,5]] inserts src[0,1,2] into dest[1,3,5].
+    slice_args : list of [src_slice_args, dest_slice_args] pairs, optional
+        Each pair defines a slice in the input and output vectors.
+        Example: slice_args=[[0,3], [2,5]] inserts src[0:3] into dest[2:5].
+        Multiple pairs: slice_args=[[0,2],[0,2]], [[2,4],[5,7]]]
```
alfiopuglisi: Missing opening square bracket

guidoagapito: Done

alfiopuglisi (approved):

PR #511, opened 2026-03-03
Zwfs_slopes

Menessao:
Added zwfs_slopec for the computation of slopes (meta-intensities) with the Zernike WFS.

An example use of the class in the configuration file is as follows:
zwfs_slopes:
  class:             'ZwfsSlopec'
  diameter:          20
  ccd_size:          [40,40]
  obsratio:          0.0
  inputs:
    in_pixels:        'ocam2.out_pixels'
  outputs:           ['out_slopes']

I am not convinced by the naming (diameter should probably be pup_diameter and ccd_size might become pix_size), but this seems to work, even with slope nulls

Additional changes:
- Pupdata class was modified to accomodate an arbitrary number of pupils
- Slopes class was modified to allow for an odd number of slopes
- constant_max and constant_min operators added to base_operation


specula/data_objects/pupdata.py
```diff
@@ -57,13 +59,20 @@ def set_value(self, v):
...
+    def n_pupils(self):
+        return self.ind_pup.shape[1]
 
     def pupil_idx(self, n):
         return self.ind_pup[:, n]
 
     def zcorrection(self, indpup):
         tmp = indpup.copy()
-        tmp[:, 2], tmp[:, 3] = indpup[:, 3], indpup[:, 2]
+        if tmp.shape[1] == 3:
```
alfiopuglisi: Shouldn't this be 4 instead of 3?

specula/data_objects/pupdata.py
```diff
@@ -76,12 +85,11 @@ def display_map(self):
...
-                self.pupil_idx(0)[self.pupil_idx(0) >= 0],  # A
-                self.pupil_idx(1)[self.pupil_idx(1) >= 0],  # B  
-                self.pupil_idx(2)[self.pupil_idx(2) >= 0],  # C
-                self.pupil_idx(3)[self.pupil_idx(3) >= 0]   # D
-            ])
+            #     self.pupil_idx(0)[self.pupil_idx(0) >= 0],  # A
+            #     self.pupil_idx(1)[self.pupil_idx(1) >= 0],  # B  
+            #     self.pupil_idx(2)[self.pupil_idx(2) >= 0],  # C
+            #     self.pupil_idx(3)[self.pupil_idx(3) >= 0]   # D
+            return self.xp.concatenate([self.pupil_idx(i)[self.pupil_idx(i) >= 0] for i in range(1)])
```
alfiopuglisi: I guess the for loop is necessary to generalize to an arbitrary number of pupils? Anyway the "for i in range(1)" looks wrong, it should be "for i in range(self.n_pupils)"

specula/data_objects/pupdata.py
```diff
@@ -94,7 +102,7 @@ def single_mask(self):
 
     def complete_mask(self):
         f = self.xp.zeros(self.framesize, dtype=self.dtype)
-        for i in range(4):
+        for i in range(self.n_pupils()):
```
alfiopuglisi: n_pupils is a property so there should be no need for parenthesis

alfiopuglisi (changes requested):

alfiopuglisi (approved):

guidoagapito (approved):

PR #519, opened 2026-03-04
(FRO) Adding docstrings to some processing objects.

FabioRossiArcetri:


specula/processing_objects/sn_calibrator.py
```diff
@@ -14,6 +14,12 @@ def __init__(self,
                  target_device_idx: int = None,
                  precision: int = None
                 ):
+        """
```
guidoagapito: This should be moved after the class definition as in the other processing object.

alfiopuglisi (approved):

guidoagapito (approved):

PR #520, opened 2026-03-05
(GA) automated generation of object summary in documentaion

guidoagapito:


docs/scripts/generate_objects_summary.py
```diff
@@ -0,0 +1,136 @@
...
+
+def generate_rst_table(category_name, modules, description=''):
+    """Generate RST content with a table listing class names and short descriptions."""
+    # Pre-count valid classes
+    total = sum(
+        1
+        for _, filepath in modules
+        for classname in get_class_short_doc(filepath)
+        if not classname.startswith('_')
+    )
```
alfiopuglisi, line 61: This iteration to count classes is exactly the same as done later in line 79 to generate the summaries, so get_class_short_doc() is called twice for all classes.
Instead of doing sum(1 for ..), it could extract all the valid class names already (and concatenating module_name.class_name into a single string to that a simple dict can be used). So the total just become len(valid_classes) and the next iteration in line 79 is just "for name, short_doc in valid_classes"

guidoagapito: You are right, see https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/522

PR #527, opened 2026-03-06
Added infinite_loop flag to LoopControl and main YAML parameters

alfiopuglisi:


guidoagapito (approved):

FabioRossiArcetri (commented):
what do you think about having this functionality simply when if the duration of the simulation is set to -1?

alfiopuglisi:
> what do you think about having this functionality simply when if the duration of the simulation is set to -1?

Right, it is easier with -1, I have updated the PR

FabioRossiArcetri (approved):

PR #536, opened 2026-03-09
Zernike fix

Menessao:
Debugging of the zernike_sensor.py class

Added the flux_inside_ccd attribute to the ModulatedPyramid class to account for the flux lost when using a detector smaller than fft_res x pup_diam 

Also added the get_psf_profile method to the PSF class, using the computeRadialProfile function

### DIFF
--- a/specula/processing_objects/modulated_pyramid.py
+++ b/specula/processing_objects/modulated_pyramid.py
@@ -265,7 +265,7 @@ class ModulatedPyramid(BaseProcessingObj):
         self.transmission = BaseValue(value=self.xp.zeros(1, dtype=self.dtype),
                                       target_device_idx=self.target_device_idx,
                                       precision=precision)        
-        self.flux_inside_ccd = BaseValue(value=self.xp.ones(1, dtype=self.dtype),
+        self.flux_frac_inside_ccd = BaseValue(value=self.xp.ones(1, dtype=self.dtype),
                                       target_device_idx=self.target_device_idx,
                                       precision=precision)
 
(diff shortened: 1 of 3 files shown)


specula/processing_objects/modulated_pyramid.py
```diff
@@ -271,6 +274,7 @@ def __init__(self,
         self.outputs['out_psf_tot'] = self.psf_tot
         self.outputs['out_psf_bfm'] = self.psf_bfm
         self.outputs['out_transmission'] = self.transmission
+        self.outputs['out_flux_inside_detector'] = self.flux_inside_ccd
```
guidoagapito: This is a fraction of flux, because it is normalized by self.xp.sum(ccd_internal), so I suggest to change names:
- self.flux_frac_inside_ccd
- out_flux_frac_inside_detector

specula/processing_objects/zernike_sensor.py
```diff
@@ -14,14 +14,14 @@ def __init__(self,
...
-                 fft_res: float = 8.0,
+                 spot_radius_lambda: float= 1.0,  # Spot radius in λ/D units
+                 phase_shift_pi: float = 0.5,  # π/2 phase shift
+                 fft_res: float = 4.0,
                  target_device_idx=None,
                  precision=None):
 
         self.spot_radius_lambda = spot_radius_lambda
-        self.phase_shift = phase_shift
+        self.phase_shift = phase_shift_pi
```
alfiopuglisi: I suggest to rename the class attribute to "phase_shift_pi" as well, if we want to keep the "pi" unit. Otherwise it might be confused with radians

alfiopuglisi (approved):

guidoagapito (approved):

PR #538, opened 2026-03-10
Preliminary interactive input

alfiopuglisi:
Adds a TerminalInput processing object.
The provided parameter file params_pyr_soul_dynamic.yml removes the PSF verbose output, in order to be able to use the terminal.

How to test on gandalf:

1) Start the simulation:
specula config/SOUL/params_pyr_soul_dynamic.yml --target=1 --no-speed-report

2) Open a browser to gandalf:8080, open the simulation display and the psf.out_sr plot

3) type the following commands on the terminal:
"gain 0.001"
"gain 0.4"
"reset"

Note that "gain 0" cannot be used due to a bug in IirFilterData


specula/simul.py
```diff
@@ -1004,7 +1004,7 @@ def run(self):
             disp.name = 'display_server'
 
         # Run simulation loop
-        self.loop.run(run_time=self.mainParams['total_time'], dt=self.mainParams['time_step'], speed_report=True)
+        self.loop.run(run_time=self.mainParams['total_time'], dt=self.mainParams['time_step'], speed_report=False)
```
guidoagapito: Could this argument be made configurable in either the yaml file or the Specula command?

alfiopuglisi: Yes, I added it as a command line argument

guidoagapito (approved):

specula/processing_objects/integrator.py
```diff
@@ -31,24 +37,51 @@ def __init__(self,
...
+        if n_modes is None or array_to_repeat is None:
+            return array_to_repeat
+        if type(n_modes) is list and (
+                type(array_to_repeat) is not list or
+                len(n_modes) != len(array_to_repeat)):
+            raise ValueError(f"When n_modes is a list, length of n_modes {len(n_modes)} must"
+                             f" match length of {array_name} {len(array_to_repeat)}")
+        return np.repeat(array_to_repeat, n_modes)
+
+    def prepare_trigger(self, t):
```
guidoagapito: This could be moved to the IirFilter class because it can be applied to any IIR filter.

config/SOUL/params_pyr_soul_gain.yml
```diff
@@ -0,0 +1,167 @@
...
+  pixel_pupil:       220                    # Linear dimension of pupil phase array
+  pixel_pitch:       0.037372727            # [m] Pitch of the pupil phase array
+  total_time:        30.00                # [s] Total simulation running time
+  time_step:         0.000588               # [s] Simulation time step
+  zenithAngleInDeg:  0.000
+  display_server:    True
+
+input:
+  class:             'TerminalInput'
+  output_list:       ['gain', 'reset']
```
guidoagapito, line 15: If we have more than one TerminalInput is present we should return an error if the same string in output_list is used in more than one TerminalInput.

specula/simul.py
```diff
@@ -54,7 +54,8 @@ def __init__(self,
                  diagram=False,
                  diagram_title=None,
                  diagram_filename=None,
-                 diagram_colors_on=False
+                 diagram_colors_on=False,
+                 speed_report=False,
```
guidoagapito: Speed report could be True by default and disabled with this argument or when a TerminalInput is present in the parameters file.

guidoagapito (changes requested):

config/SOUL/params_pyr_soul_gain.yml
alfiopuglisi: Only a single TerminalInput can be used, since multiple instances cannot share the same terminal. I have added a check during instantiation

specula/processing_objects/integrator.py
alfiopuglisi: done

specula/simul.py
alfiopuglisi: I have reversed the command line option, now by default the speed report is present and one can use --no-speed-report to disable it when using the terminal. I would prefer not to change parameters automatically.

guidoagapito (approved):

PR #539, opened 2026-03-13
Reorganized display API

alfiopuglisi:


config/scao/calib_pyr_pupdata_display.yml
```diff
@@ -0,0 +1,30 @@
+
+
+pupil_calibrator:
```
guidoagapito: I think this file does not belong to this pull request

alfiopuglisi: Right, wrong branch, I will remove it

FabioRossiArcetri (approved):

guidoagapito (approved):

guidoagapito:
chatGPT propose to fix the test as:

```
diff --git a/test/test_display_server.py b/test/test_display_server.py
index f954d07..446c1fc 100644
--- a/test/test_display_server.py
+++ b/test/test_display_server.py
@@ -318,13 +318,21 @@ class TestDisplayServerTrigger(unittest.TestCase):
         server.qin.put(('client_xyz', ['obj1']))
 
         server.trigger()
-        time.sleep(0.001) # Queue context switch
+        time.sleep(0.001)  # Queue context switch
 
-        items = []
-        while not server.qout.empty():
-            items.append(server.qout.get_nowait())
+        types = []
+        deadline = time.time() + 0.5
+        while time.time() < deadline:
+            try:
+                item = server.qout.get(timeout=0.05)
+            except _queue_module.Empty:
+                continue
+
+            if isinstance(item, tuple):
+                types.append(item[0])
+                if item[0] == 'terminator':
+                    break
 
-        types = [i[0] for i in items if isinstance(i, tuple)]
         self.assertIn('terminator', types)
```

This fixes a CI-only flaky test in the display server trigger path.
The previous assertion used `multiprocessing.Queue.empty()`+ `get_nowait()`, which is timing-dependent and can miss queued items on slower runners, so the test sometimes saw only the status message (sim) and not the terminator.
The test now polls the queue with a small timeout and waits until terminator is observed (or timeout), making the assertion deterministic without changing production behavior.

PR #544, opened 2026-03-16
(GA) chromatic effects in AtmoPropagation with model for the refractive index and tests

guidoagapito:


specula/processing_objects/atmo_propagation.py
```diff
@@ -25,6 +26,9 @@ def __init__(self,
                  source_dict: dict,     # TODO ={},
                  doFresnel: bool=False,
                  wavelengthInNm: float=500.0,
+                 telescope_altitude_m: float=3064.0,
```
alfiopuglisi: Isn't it better to set the altitude default to None as well, like the reference wavelength, so that one has to make sure it is for the correct telescope? Otherwise 99% of simulations will use the ELT altitude for sure

alfiopuglisi (approved):
See inline comments

specula/processing_objects/atmo_propagation.py
guidoagapito: Ok, good idea!

Done

PR #546, opened 2026-03-17
Lift implementation

FabioRossiArcetri:


specula/processing_objects/lift.py
```diff
@@ -0,0 +1,477 @@
...
+'''
+
+class Lift(BaseProcessingObj):
+    """
+    SPECULA ProcessingObject implementing the LIFT algorithm    
+    """
+
+    def __init__(self,
+                 simul_params: SimulParams,                 
+                 defocus_amp,
```
alfiopuglisi: Add types to arguments if possible?

specula/processing_objects/lift.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def trigger(self):
+        if self.in_pixels is None:
+            return
+
+        psf = self.in_pixels.get_value()
+        currentPhaseEstimate, coeffs, niters = self.phaseEstimation(psf)        
+        
+        self.outputs['out_modes'].value = self.to_xp(coeffs)
+        self.outputs['out_modes_0'].value = coeffs[0]
```
alfiopuglisi: Is the purpose of the 5 outputs to have the history? If so, I am not sure it works since "coeffs" array is just the last estimate

specula/processing_objects/lift.py
```diff
@@ -0,0 +1,477 @@
...
+                 target_device_idx: int = None,
+                 precision: int = None):
+
+        super().__init__(target_device_idx=target_device_idx, precision=precision)
+
+        self.simul_params = simul_params
+        # Store parameters
+        self.nPistons = nPistons
+        self.nZern = nZern
+        self.airef = np.zeros(nPistons + nZern)
```
alfiopuglisi: dtype is missing

specula/processing_objects/lift.py
```diff
@@ -0,0 +1,477 @@
...
+        self.fix = bool(fix)
+
+        # Derived parameters
+        self.nmodes = nZern + nPistons
+        self.modes = None
+        self.phase_ref = None
+        self.ref_tip = 0.0
+        self.ref_tilt = 0.0
+        self.padded = None
+        self.radians_per_pixel = np.pi / (2 * 3)
```
alfiopuglisi: Wasn't radians_per_pixel calculated based on the gridsize?

alfiopuglisi (commented):
See inline comments

specula/processing_objects/lift.py
guidoagapito: I don't think this five-output model is general, but rather valid for cases where estimation is performed on five modes.

specula/processing_objects/lift.py
guidoagapito: Is np selected by purpose?

specula/processing_objects/lift.py
alfiopuglisi: Then it should be a single output with a vector, because the number of modes is configurable with the nZern parameter

specula/processing_objects/lift.py
guidoagapito: This variable converts the spatial displacement of the PSF (in pixels) into a Tip/Tilt phase coefficient (in radians) to provide an initial guess for the phase estimation. By hardcoding it to pi / 6, we are implicitly assuming a fixed oversampling factor of 1.5 (since the IDL PASSATA equivalent calculates this dynamically as !pi/(4.*self._oversamp)).

If a user changes the simulation geometry parameters (like fft_res, pix_scale, or npix_side), this hardcoded conversion factor will yield an inaccurate initial Tip/Tilt estimate. While the iterative algorithm might still eventually converge, an inaccurate initial guess forces it to perform unnecessary iterations, slowing down the real-time trigger() loop.

I propose to set this variable in set_modalbase after calc_geometry.
self.radians_per_pixel = np.pi / (4.0 * settings.fft_res)

specula/processing_objects/lift.py
```diff
@@ -0,0 +1,477 @@
...
+                logging.debug(f"center_x, center_y: {center[0]}, {center[1]}")
+                logging.debug(f"min R, max R: {Rdiag.min()}, {Rdiag.max()}")
+                logging.debug(f"Current step of modal coefficients: {DeltaZ}")
+                logging.debug(f"Current estimate of modal coefficients: {total_A_ML}")
+                logging.debug(f"Criterion value: {newerror}")
+            if abs(improvement) < relTol or np.mean(np.abs(DeltaZ)) < absTol:
+                logging.debug(f'criterion reached at the {i+1} iteration')
+                break
+
+        total_A_MLs[-1][0+self.nPistons] += self.ref_tip - 0.5 * self.radians_per_pixel
```
guidoagapito: The final modal coefficients are adjusted with a - 0.5 * self.radians_per_pixel offset.

This adjustment is acting as a patch for how the geometric center of the grid is calculated initially. When setting the reference Tip/Tilt (setRefTT), the code uses integer division N // 2 to define the center of the array.

However, for an array of size N, the true mathematical center in float coordinates is exactly (N - 1) / 2.0.

specula/processing_objects/lift.py
```diff
@@ -0,0 +1,477 @@
...
+        total_A_MLs = []
+        errors.append(olderror)
+        currentPhaseEstimates.append(currentPhaseEstimate)
+        total_A_MLs.append(total_A_ML)
+        # Estimate initial ROI and flux
+        center = self.calcCenter(psf)
+        flux = self.calcCroppedFlux(psf, center)
+        # Set reference TT based on initial ROI
+        N = psf.shape[0]
+        self.setRefTT((center[0] - N//2)*self.radians_per_pixel,
```
guidoagapito: see comment on "- 0.5 * self.radians_per_pixel"

guidoagapito:
A general comment: I suggest to add some tests.

alfiopuglisi (approved):

specula/processing_objects/lift.py
```diff
@@ -0,0 +1,432 @@
...
+
+        self.inputs['in_pixels'] = InputValue(type=Pixels)
+
+        self.out_modes = BaseValue(
+            value=self.xp.zeros(self.nmodes, dtype=self.dtype),
+            target_device_idx=target_device_idx
+        )
+        self.out_modes.value = self.xp.zeros(self.nmodes, dtype=self.dtype)
+
+        self.outputs['out_modes'] = self.out_modes
```
alfiopuglisi: Just an idea, could it be useful to have two separate outputs: one for pistons, and another for low-order modes?

guidoagapito: I separated the pistons and Zernikes.
I also made some other changes.

alfiopuglisi (approved):

PR #547, opened 2026-03-18
(GA) complex number dtype

guidoagapito:


specula/processing_objects/four_quadrant_coronagraph.py
```diff
@@ -48,7 +48,9 @@ def make_focal_plane_mask(self):
                                 yc=1.0+1./self.fft_totsize, xp=self.xp,
                                 square=True, xc = 1./self.fft_totsize)
         quad_mask = self.xp.logical_xor(left_mask,bottom_mask)
-        fp_mask = self.xp.exp(1j*quad_mask*self._phase_delay, dtype=self.complex_dtype)
+        iu = self.complex_dtype(1j)
+        phase_delay = self.dtype(self._phase_delay)
+        fp_mask = self.xp.exp(iu * quad_mask * phase_delay, dtype=self.complex_dtype)
```
alfiopuglisi, line 53: The dtype inside exp() should do the conversion automatically, without the need to convert phase_delay and iu. Wasn't this the case?

specula/processing_objects/gain_optimizer.py
```diff
@@ -338,8 +338,9 @@ def _calculate_rejection_tf_cached(self, freq_tuple, t_int, gain, num_tuple, den
         den = self.to_xp(den_tuple, dtype=self.dtype)
 
         # Calculate transfer function
-        omega = 2 * np.pi * freq * t_int
-        z = self.xp.exp(1j * omega)
+        omega = 2 * self.dtype(np.pi) * freq * t_int
+        iu = self.complex_dtype(1j)
+        z = self.xp.exp(iu * omega, dtype=self.complex_dtype)
```
alfiopuglisi, line 343: Same comment as four_quadrant_coronograph

alfiopuglisi (commented):
See inline comments

guidoagapito:
> See inline comments

The conversion avoid performing the product in double precision and converting it to single precision in the exp.

alfiopuglisi:
> > See inline comments
> 
> The conversion avoid performing the product in double precision and converting it to single precision in the exp.

ok right, we can keep it

alfiopuglisi (approved):

PR #551, opened 2026-03-19
(GA) downsampling in data store with test

guidoagapito:


docs/tutorials/scao_basic_tutorial.rst
```diff
@@ -282,6 +282,26 @@ Create a YAML configuration file, for example ``params_scao_pyr_basic.yml``:
...
 
+If the full-rate data volume is too large, ``DataStore`` can decimate what is
+written to disk for selected inputs:
+
+.. code-block:: yaml
+
+   data_store:
+     class:             'DataStore'
+     store_dir:         './output'
+     store_every_by_input:
```
alfiopuglisi: Is this a better name than "decimation" ?

specula/processing_objects/data_store.py
```diff
@@ -61,7 +127,10 @@ def save_pickle(self):
                     continue
 
                 filename = os.path.join(self.tn_dir, k + '.pickle')
-                hdr = self.inputs[k].get(target_device_idx=-1).get_fits_header()
+                hdr = self._header_with_storage_metadata(
+                    k,
+                    self.inputs[k].get(target_device_idx=-1).get_fits_header()
+                )
```
alfiopuglisi: I find this style difficult to understand. I would be simpler to keep get_fits_header() like before, and have an update_header( ) function that adds the two new keywords (and does not need to do header.copy()). Something like:

hdr = self.inputs[k].get(target_device_idx=-1).get_fits_header()
self.update_header_with_storage_metadata(hdr, input_name=k)

alfiopuglisi (changes requested):
See inline comments

docs/tutorials/scao_basic_tutorial.rst
guidoagapito: I changed "decimation" in "downsampling"

specula/processing_objects/data_store.py
guidoagapito: Done.

guidoagapito:
> See inline comments

I implemented the suggested changes.

docs/tutorials/scao_basic_tutorial.rst
alfiopuglisi: I meant that "every_by" could be changed to "decimation"

Anyway it has little importance.

alfiopuglisi (approved):

docs/tutorials/scao_basic_tutorial.rst
guidoagapito: Ok, sorry I had misunderstood. However, I thought it would be best to use consistent names for the arguments.

Issue #380, opened 2025-09-17
Wrong error message for non-existing params


alfiopuglisi:
In some cases, if a non-existing parameter is given in the YML file for an object, the error message is very misleading:

raise ValueError(f'No type hint for parameter {parname} of class {classname}')

this happens for example if "pupdata_object" is given to ShSlopec, which does not have this parameter.

alfiopuglisi:
Fixed in #553

PR #553, opened 2026-03-20
Fix for issue #380 (wrong error message during initialization)

alfiopuglisi:


alfiopuglisi:
@guidoagapito I had to add a special case for iir_filter_data in class IirFillter: using a name ending in "_data" conflicts with our parameter name processing (like "_object", "_dict_ref", etc). Unfortunately changing this iir_filter_data looks very invasive, but it is something to keep an eye on.

specula/simul.py
```diff
@@ -364,11 +364,21 @@ def build_objects(self, params):
                 if name in skip_pars and name not in args:
                     continue
 
+                # Check that each parameter name is expected by the constructor of the class, after removing possible suffixes
+                parname = name
+                endings = ['_ref', '_data', '_object']
+                excluded = ['iir_filter_data']  # Special case for iir_filter_data, which ends in '_data'
```
guidoagapito: you should add ssr_filter_data too, but the name of the class is IirFilterData and SsrFilterData, without _. Only the file name has _: iir_filter_data.py. Did I miss something?

guidoagapito:
> @guidoagapito I had to add a special case for iir_filter_data in class IirFillter: using a name ending in "_data" conflicts with our parameter name processing (like "_object", "_dict_ref", etc). Unfortunately changing this iir_filter_data looks very invasive, but it is something to keep an eye on.

See in line comment.

specula/simul.py
alfiopuglisi: The problem is the parameter name in the constructor of IirFilter and SsrFilter: "iir_filter_data: IirFilterData", the ending in "_data" is reserved. In our Yaml files is always "iir_filter_data_ref" so the problem does not appear.

guidoagapito: OK, can we change the file names of iir_filter_data.py and ssr_filter_data.py while keeping the same class names? This would minimise the impact of the change.

alfiopuglisi: The class names are fine. But also changing "just" the parameter name is a rather big change, because it impacts the Yaml files

guidoagapito: ok, I understand.

guidoagapito (approved):

PR #554, opened 2026-03-20
Added phase screen cube object to inject user-defined phase screens

cedricplantet:


alfiopuglisi (approved):
The filename as a parameter is a bit strange with respect to the other processing objects, but the real reason is that we are missing a Phasescreen data object. It would be a rather big change, so let's keep it as it is for now

specula/processing_objects/phase_screen_cube.py
```diff
@@ -0,0 +1,132 @@
...
+
+class PhaseScreenCube(BaseProcessingObj):
+    """
+    User-defined phase screen cube data object.
+    Reads a phase screen cube from a FITS file and applies it on the specified line of sight.
+    The cube's temporal sampling does not need to match the simulation's sampling.
+    """
+    def __init__(self,
+                 simul_params: SimulParams,
+                 file_name: str,
```
guidoagapito, line 21: We should add a specific data object for this case.

I propose Spatio-temporal array, abbreviated in spatio_temp_array.py SpatioTempArray.

guidoagapito: See #559

PR #557, opened 2026-03-20
Rowcols

alfiopuglisi:
We never noticed these bugs because we always use square arrays, but we need to enforce the standard convention of x=horizontal and y=vertical, while the numpy shape is (rows, cols), therefore:

shape == (y, x)

Even better would be to get rid of x/y and always use shapes, but that's a much bigger change...


guidoagapito (approved):
I approved this pull request.
I think it would be useful to carry out some tests to verify this.

alfiopuglisi:
I added several tests in addition to the ones that were already there

PR #559, opened 2026-03-21
(GA) SpatioTempArray data object

guidoagapito:


alfiopuglisi:
This new data object is very similar to arte's TimeSeries object.
We deciced not to have arte as a dependency, should we revise that decision?

guidoagapito:
> This new data object is very similar to arte's TimeSeries object. We deciced not to have arte as a dependency, should we revise that decision?

It seems to me that the two tools complement each other: arte TimeSeries was designed for post-processing analysis of telemetry data, while SPECULA SpatioTempArray is designed for handling spatio-temporal data in a simulation-oriented structure.
If the timeSeries functionality were required in the future, we could implement a similar adapter to the one that adapts timeSeries to ERIS or other AO instruments.
In that case, arte would either be an optional dependency or solely a dependency of the data reduction module.

cedricplantet (approved):

alfiopuglisi (approved):

PR #563, opened 2026-03-26
Allow trigger when all inputs are optional

alfiopuglisi:
Currently, an object with optional inputs that are disconnected will never trigger.
This PR adds code to check whether all inputs of a processing object are optional and, if they are all disconnected, will force the trigger to be executed.
This can happen with object that produce data without needing an input, like a CCD laboratory device, that optionally have a sync input to trigger only at specific times. If this optional input is not given, the object must always trigger.

### DIFF
--- a/specula/base_processing_obj.py
+++ b/specula/base_processing_obj.py
@@ -55,9 +55,21 @@ class BaseProcessingObj(BaseTimeObj):
         self.remote_outputs[name].append(remote_output)
 
     def checkInputTimes(self):
+        '''
+        Determine whether this processing object needs to execute
+        the trigger method, based on the input states
+        '''
+        # No inputs: always trigger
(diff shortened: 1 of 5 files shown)


specula/base_processing_obj.py
```diff
@@ -55,9 +55,21 @@ def addRemoteOutput(self, name, remote_output):
...
 
     def checkInputTimes(self):
+        '''
+        Determine whether this processing object needs to execute
+        the trigger method, based on the input states
+        '''
+        # No inputs: always trigger
         if len(self.inputs)==0:
             return True
-        self.get_all_inputs()
```
guidoagapito, line 60: why was this removed?

guidoagapito (approved):

PR #565, opened 2026-03-26
(GA) calibration manager subdirs updated and test

guidoagapito:


alfiopuglisi:
We already have test_data_objects.py where the list of data objects is built as well, and maybe it's a bit nicer there, without having to use __ file__ and similar. Can we enumerate classes in the same way here, or make a common function?

guidoagapito:
> We already have test_data_objects.py where the list of data objects is built as well, and maybe it's a bit nicer there, without having to use __ file__ and similar. Can we enumerate classes in the same way here, or make a common function?

I added iter_data_object_classes in specula_testlib

alfiopuglisi (approved):

PR #567, opened 2026-03-26
(GA) fix to FieldAnalyser and precision storage in DataStore with tests

guidoagapito:


guidoagapito:
It has now been tested on Gandalf and should work as expected.
I have also added precision storage because I got bad results when I made a replay of a single precision simulation with double precision, and it took me some time to understand what was going on.

specula/processing_objects/data_source.py
```diff
@@ -17,13 +17,15 @@ class DataSource(BaseProcessingObj):
...
-                data_format: str='fits'):
+                data_format: str='fits',
+                global_precision: int=None):
         super().__init__()
         self.items = {}
         self.storage = {}
         self.data_filename = ''
         self.tn_dir = store_dir
         self.data_format = data_format
+        self.global_precision = global_precision
```
alfiopuglisi, line 28: Is this the same as the "global_precision" defined in specula/__init__.py , or something else?

guidoagapito: Yes, it is the same precision set up in the specula.init(...)

alfiopuglisi (approved):

PR #569, opened 2026-03-27
Dark calibrator

alfiopuglisi:
Tested in the cascading lab


specula/processing_objects/dynamic_dark_calibrator.py
```diff
@@ -0,0 +1,144 @@
+import os
+from specula.base_processing_obj import BaseProcessingObj
+from specula.base_value import BaseValue
+from specula.data_objects.pixels import Pixels
+from specula.connections import InputValue
+from specula.data_objects.simul_params import SimulParams
+from specula.lib import utils
+
+
+class DynamicDarkCalibrator(BaseProcessingObj):
```
guidoagapito, line 10: It seems to me that this class has two features: (1) calibrating the dark frame and (2) subtracting the dark frame from the pixels.
If so, I would add a brief description to the docstring.
I also see that dark calibration depends on the nframes variable. I think it would be useful to include a brief description of this argument too.

In general, and perhaps for a different pull request, we need to document the purpose of these dynamic processing objects.

guidoagapito (approved):
I approved the pull request and I suggested "in line" to add a few comments.

PR #571, opened 2026-03-29
Minor bugfixing

Menessao:
- Possibility to repeat the entire push-pull time history: this allows to calibrate the IM in partial correction only updating the random phase screens once a calibration is completed (i.e. calibrate all the modes on the same phase-screens, rather than only guaranteeing the same phase screen for the push-pull of each mode)

- Allow to override nested inputs (e.g. data_store.input_list.inputs)

- Prevent data-store from failing if create_tn is False and the directory does not exist (it now makes the directory, can be changed to raise the appropriate error)


Menessao:
I have to figure out how to avoid having all the previous (outdated/already merged) commits showing in each PR

specula/lib/modal_pushpull_signal.py
```diff
@@ -16,6 +16,7 @@ def modal_pushpull_signal(
     ncycles: int = 1,
     repeat_ncycles: bool = False,
     nsamples: int = 1,
+    ntiles: int = 1,
```
guidoagapito: Do we need an additional int argument?
I think we need either the xp.tile or the xp.repeat, but not both.
If so, we could have a boolean argument like 'repeat_full_sequence' and use ncycles instead of ntiles.

specula/simul.py
```diff
@@ -780,9 +780,14 @@ def apply_overrides(self, params):
...
             for k, v in yaml.full_load(self.overrides).items():
-                obj_name, param_name = k.split('.')
-                params[obj_name][param_name] = v
-                print(obj_name, param_name, v)
+                try:
+                    obj_name, param_name = k.split('.')
+                    params[obj_name][param_name] = v
+                    print(obj_name, param_name, v)
+                except ValueError:
+                    obj_name, param_name, subparam_name = k.split('.')
```
guidoagapito: I think we should check the length of k.split('.')

```
def apply_overrides(self, params):
        print('overrides:', self.overrides)
        if len(self.overrides) > 0:
            for k, v in yaml.full_load(self.overrides).items():
                parts = k.split('.')
                
                if len(parts) == 2:
                    params[parts[0]][parts[1]] = v
                    print(*parts, v)
                elif len(parts) == 3:
                    params[parts[0]][parts[1]][parts[2]] = v
                    print(*parts, v)
                else:
                    raise ...
```

guidoagapito:
> I have to figure out how to avoid having all the previous (outdated/already merged) commits showing in each PR

Don't worry about this, what's important are the changed files. Squash and merge would combine all the changes into a single commit.

specula/simul.py
```diff
@@ -780,9 +780,15 @@ def apply_overrides(self, params):
...
         if len(self.overrides) > 0:
             for k, v in yaml.full_load(self.overrides).items():
-                obj_name, param_name = k.split('.')
-                params[obj_name][param_name] = v
-                print(obj_name, param_name, v)
+                parts = k.split('.')
+                if len(parts) == 2:
+                    params[parts[0]][parts[1]] = v
+                    print(*parts, v)
+                elif len(parts) == 3:
```
guidoagapito, line 787: Please add a test for this case in test_simul.py

specula/lib/modal_pushpull_signal.py
```diff
@@ -16,6 +16,7 @@ def modal_pushpull_signal(
     ncycles: int = 1,
     repeat_ncycles: bool = False,
     nsamples: int = 1,
+    repeat_full_sequence: bool = False,
```
guidoagapito, line 19: Please add a test for this argument in test_modal_pushpull_signal.py

FabioRossiArcetri (approved):

specula/processing_objects/push_pull_generator.py
```diff
@@ -17,7 +17,7 @@ def __init__(self,
                  vect_amplitude: list = None,
                  ncycles: int = 1,
                  nsamples: int = 1,
-                 repeat_cycles: bool = False,
+                 repeat_full_sequence: bool = False,
```
guidoagapito, line 21: I would like a different implementation of this feature.

1. Separation of concerns: nsamples vs. pattern repetition
It was previously noted that repeat_ncycles might be redundant because one could just increase nsamples. While mathematically repeat_ncycles=True with ncycles=3 yields the same flat array as ncycles=1 and nsamples=3, they represent two fundamentally different concepts. nsamples defines the hold time (simulation frames per command), whereas repeating the cycles defines the actuation sequence logic.
If a simulation requires 3 consecutive pushes, and each command must be held for 5 simulation frames, asking the user to set nsamples=15 breaks the semantic meaning of the parameter. Keeping a boolean flag for the actuation logic strictly separates the sequence generation from the temporal sampling.

2. Keep both boolean flags to avoid losing functionality
Because of the reason above, removing repeat_ncycles and replacing it entirely with repeat_full_sequence removes a valid use case (grouped sequence for a single mode, e.g., + + - -).
A more flexible approach is to keep one scalar (ncycles) and two independent boolean flags (repeat_ncycles and repeat_full_sequence). This covers all use cases without cluttering the function signature.

3. Optimize memory allocation
Currently, time_hist is allocated using * ncycles right away. If repeat_full_sequence is True, the for loop only fills a small portion of this large array. Then, the code slices it (time_hist[:n_pokes * real_n_modes,:]) and uses xp.tile to allocate yet another large array. For a high number of modes or cycles, especially when using CuPy on GPUs, this allocates a lot of redundant memory for nothing.

We can fix all these issues by dynamically setting the local_cycles and allocating only the strictly needed memory before tiling. Here is a proposed snippet for the core loop:

    # 1. Decide local cycles based on the requested behavior
    local_cycles = 1 if repeat_full_sequence else ncycles
    
    real_n_modes = n_modes - first_mode
    
    # 2. Allocate ONLY the strictly necessary memory
    time_hist = xp.zeros((n_pokes * real_n_modes * local_cycles, n_modes))
    
    for mode in range(first_mode, n_modes):
        hist_idx = mode - first_mode
        poke_pattern = vect_amplitude[mode] * xp.array(pattern)
        
        # 3. Support both local repetition (+ + - -) and alternating (+ - + -)
        if repeat_ncycles and not repeat_full_sequence:
            time_hist[n_pokes*hist_idx*local_cycles:n_pokes*(hist_idx+1)*local_cycles, mode] = \
                xp.repeat(poke_pattern, local_cycles)
        else:
            for j in range(local_cycles):
                time_hist[n_pokes*(local_cycles*hist_idx+j):n_pokes*(local_cycles*hist_idx+j+1), mode] = poke_pattern
    
    # 4. Repeat the whole block if requested, without slicing empty arrays
    if repeat_full_sequence:
        time_hist = xp.tile(time_hist, [ncycles, 1])

    return xp.repeat(time_hist, nsamples, axis=0)

guidoagapito (approved):

PR #575, opened 2026-03-31
Logging

alfiopuglisi:
Using logging module instead of print() functions and verbose flags.
Most of the new logic is in specula/log.py.
All objects get a default self.logger with a pre-defined log leve, which can be overidden using a command line argument, for exampe: --log-level=debug


guidoagapito (approved):

FabioRossiArcetri:
almost there, just lacking a bit of coverage

alfiopuglisi:
> almost there, just lacking a bit of coverage

Before merging, we should agree on what to do with all the verbose flags. Convert them to log level? Decide on a specific log level for the verbose flag, and another one without?

guidoagapito:
> > almost there, just lacking a bit of coverage
> 
> Before merging, we should agree on what to do with all the verbose flags. Convert them to log level? Decide on a specific log level for the verbose flag, and another one without?

I think verbose flags should change the log level of the class.
We should also document all log levels and explain how to set them.

FabioRossiArcetri:
Having both the verbose flag and the log level seams redundant. I would go for removing verbose and replacing it with log level: unfortunately this would change the interface, but it is a consequence of the change in the philosophy of info/debug messages. An alternative, which would be keeping the verbose parameter and transforming it into an int debug level seams misleading.

specula/log.py
```diff
@@ -0,0 +1,87 @@
...
+
+    logging.basicConfig(level=log_level, format=log_format)
+
+    # Make sure all loggers use our filter
+    root = logging.getLogger()
+    for handler in root.handlers:
+        handler.addFilter(SpeculaFilter(process_rank))
+
+
+class SpeculaFilter(logging.Filter):
```
guidoagapito: I suggest to use a name that clearly refers to log, like SpeculaLogFilter

specula/log.py
```diff
@@ -0,0 +1,87 @@
...
+        if hasattr(record, "instance_name") and record.instance_name:
+            if record.levelno <= logging.DEBUG or record.instance_name in [INIT_PLACEHOLDER_NAME]:
+                record.name = f'{record.name} - {record.instance_name}'
+            else:
+                record.name = record.instance_name
+        record.process_rank = self.process_rank
+        return True
+
+
+class SpeculaAdapter(logging.LoggerAdapter):
```
guidoagapito: I suggest to use a name that clearly refers to log, like SpeculaLogAdapter

specula/log.py
```diff
@@ -0,0 +1,87 @@
...
+    '''
+    Logger adapter that defines custom log levels for MPI debugging, below the standard DEBUG level (10):
+    - MPI_DBG_LEVEL (6): General MPI debugging messages
+    - MPI_SEND_DBG_LEVEL (5): Detailed messages for MPI send/receive operations
+    Also manages the instance name for log records, allowing it to be included in the log output.
+    '''
+    def __init__(self, logger):
+        super().__init__(logger, {})
+
+    MPI_DBG_LEVEL = 6
```
guidoagapito: Do we need an addLevelName for these log levels?

specula/processing_objects/base_sprint_estimator.py
```diff
@@ -278,15 +273,14 @@ def setup(self):
...
-            print(f"  Number of slopes: {self.estimated_intmat.nslopes}")
-            print(f"  Size of pupil: {self.pupil_mask.shape}")
-            print(f"  Size of DM influence functions: {self.ifunc_3d.shape}")
-            print(f"  Estimation interval: {self.t_to_seconds(self.estimation_dt):.2f}s")
-            print(f"  Integration gain: {self.integration_gain}")
-            print(f"  Forgetting factor: {self.forgetting_factor}")
+        self.logger.info(f"  initialized:")
+        self.logger.info(f"  Number of modes: {self.nmodes}")
+        self.logger.info(f"  Number of slopes: {self.estimated_intmat.nslopes}")
+        self.logger.info("  Size of pupil: {self.pupil_mask.shape}")
```
guidoagapito: an f is missing

specula/base_processing_obj.py
```diff
@@ -99,29 +99,27 @@ def get_all_inputs(self):
         Data is transferred between devices if necessary.
         '''
         for input_name, input_obj in self.inputs.items():
-            if MPI_DBG: print(process_rank, 'get_all_inputs(): getting InputValue:',
-                              input_name, flush=True)
+            self.logger.mpi_debug(f'- get_all_inputs(): '
+                                    'getting InputValue: {input_name}')
```
guidoagapito: an f is missing

specula/simul.py
```diff
@@ -601,14 +607,14 @@ def build_objects(self, params):
             my_params.update(pars2)
             try:
                 self.objs[key] = klass(**my_params)
+                self.objs[key].name = key
+                self.objs[key].init_logging(self.logger.getEffectiveLevel())
             except Exception:
-                print('Exception building', key)
+                self.logger.error('Exception building {key}')
```
guidoagapito: an f is missing

specula/log.py
```diff
@@ -0,0 +1,87 @@
...
+    Initialize logging with a custom format that includes the process rank if provided,
+    and set up logging with our SpeculaFilter enabled.
+    '''
+    if log_format is None:
+        if process_rank is None:
+            log_format="%(asctime)s [%(levelname)s]: [%(name)s]: %(message)s"
+        else:
+            log_format="%(asctime)s [%(levelname)s]: [rank %(process_rank)s] [%(name)s]: %(message)s"
+
+    logging.basicConfig(level=log_level, format=log_format)
```
guidoagapito: This is a global configuration that can change the behaviour of other libraries.

specula/log.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Make sure all loggers use our filter
+    root = logging.getLogger()
+    for handler in root.handlers:
+        handler.addFilter(SpeculaFilter(process_rank))
```
guidoagapito: This is a global configuration that can change the behaviour of other libraries.

specula/log.py
```diff
@@ -0,0 +1,87 @@
...
+    Also add the process rank to the log record.
+    '''
+    def __init__(self, process_rank):
+        super().__init__()
+        self.process_rank = process_rank
+
+    def filter(self, record):
+        if hasattr(record, "instance_name") and record.instance_name:
+            if record.levelno <= logging.DEBUG or record.instance_name in [INIT_PLACEHOLDER_NAME]:
+                record.name = f'{record.name} - {record.instance_name}'
```
guidoagapito: SpeculaFilter currently mutates LogRecord.name to show the instance name (and sometimes class - instance). This is convenient for readability, but it can (a) collapse distinct emitters when instance_name is not unique/placeholder and (b) break downstream logging configs that rely on the specula.* logger hierarchy (filters/routing/levels based on record.name).

Alternative approach: keep record.name untouched and expose the instance label via an extra field (e.g. record.instance_name / record.specula_instance) and include it in the formatter ("%(name)s [%(instance_name)s]: %(message)s"), or use a LoggerAdapter + formatter to display instance info without rewriting the logger name.

test/test_base_time_obj.py
```diff
@@ -79,13 +79,13 @@ def test_start_and_stop_mem_usage_count_gpu(self, target_device_idx, xp):
     def test_print_mem_usage_calls_print_on_gpu(self, target_device_idx, xp):
         """Ensure printMemUsage prints only for GPU."""
         obj = BaseTimeObj(target_device_idx=target_device_idx)
-        with patch("builtins.print") as mock_print:
+        with patch.object(obj.logger, "log") as mock_log:
             obj.gpu_bytes_used = 1048576  # 1MB
             obj.printMemUsage()
             if target_device_idx >= 0:
-                mock_print.assert_called_once()
+                mock_log.assert_called_once()
```
guidoagapito: Some of the updated tests look a bit brittle because they assert exact logging call counts (e.g. assert_called_once() / assert_not_called()) or mock logger.log directly. Small, non-behavioral changes (extra debug/info messages, refactoring .info → .warning, etc.) will cause unnecessary failures.

Suggestion: prefer capturing log records and asserting on intent (level + key substrings) rather than call counts. In unittest, assertLogs("specula", level=...) works well; in pytest, caplog is even nicer. When you need “no logs”, consider asserting “no WARNING/ERROR emitted” for the relevant logger instead of “no calls at all”.

guidoagapito:
I added my comments, part of them are suggested by copilot.

guidoagapito:
I forgot:

I think we need some documentation: (1) what is the log level in SPECULA (2) what are the log levels (3) how to set the desired log level.

guidoagapito (changes requested):

specula/log.py
alfiopuglisi: done

specula/log.py
alfiopuglisi: done

alfiopuglisi: Done

specula/log.py
alfiopuglisi: Done, it was in fact necessary for the command-line argument to work (e.g. --log-level=mpi_dbg)

specula/processing_objects/base_sprint_estimator.py
alfiopuglisi: done

specula/base_processing_obj.py
alfiopuglisi: done

specula/simul.py
alfiopuglisi: done

test/test_base_time_obj.py
alfiopuglisi: done

specula/log.py
alfiopuglisi: Yes but from what I understand it is ok: SPECULA is the main program, and it can (and should) configure the logging behaviour. Libraries should use logging without configuring it

specula/log.py
alfiopuglisi: Same as before. In addition, SpeculaFilter only acts if "instance_name" is present, and that's a SPECULA-only feature, so it won't affect other libraries.

specula/log.py
alfiopuglisi: Done, refactored handlers and formatters

alfiopuglisi:
> I forgot:
> 
> I think we need some documentation: (1) what is the log level in SPECULA (2) what are the log levels (3) how to set the desired log level.

I added documentation in running_simulations.rst

specula/log.py
alfiopuglisi: This is partly fixed because, after the last refactor, it does not call basicConfig anymore

docs/running_simulations.rst
```diff
@@ -38,6 +38,71 @@ These arguments are similar to the ones used by ``specula`` itself, whose implem
...
 
+Output logging
+==============
+
+Simulation output is written using the standard Python ``logging`` module. As a default, it will print on standard output a timestamped informational line like:
+
+   2026-04-19 14:18:45,206 [INFO]: [psf]: SR at 1650nm : 0.0033
+
+Each processing object handles its output and defines what to print, and at which log level. The log level for the whole simulation
+is set with the ``--log-evel`` command line switch: this option sets the minimum severity level of log messages that will be emitted.
```
guidoagapito: Typo, --log-evel

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #577, opened 2026-03-31
Added input_names and output_names

alfiopuglisi:
PR to show changes

### DIFF
--- a/specula/processing_objects/ext_source_pyramid.py
+++ b/specula/processing_objects/ext_source_pyramid.py
@@ -216,11 +216,12 @@ class ExtSourcePyramid(ModulatedPyramid):
 
     @staticmethod
     def input_names():
-        return ModulatedPyramid.input_names().update( {
+        input_names = ModulatedPyramid.input_names()
+        input_names.update({
             'ext_source_coeff': (BaseValue, 'Extended source coefficients array of shape (n_points, 4)'
                                             ' with columns: [tip_coeff, tilt_coeff, focus_coeff, flux]')
-        } )
-
+        })
+        return input_names
 
     def _get_pyramid_face_angles_at_fov_radius(self):
         """
--- a/specula/processing_objects/ext_source_pyramid.py
+++ b/specula/processing_objects/ext_source_pyramid.py
@@ -214,6 +214,13 @@ class ExtSourcePyramid(ModulatedPyramid):
         # Add dedicated input for extended source coefficients
         self.inputs['ext_source_coeff'] = InputValue(type=BaseValue)
 
+    @staticmethod
+    def input_names():
+        return ModulatedPyramid.input_names() | {
+            'ext_source_coeff': (BaseValue, 'Extended source coefficients array of shape (n_points, 4)'
+                                            ' with columns: [tip_coeff, tilt_coeff, focus_coeff, flux]')
+        }        
+
 
     def _get_pyramid_face_angles_at_fov_radius(self):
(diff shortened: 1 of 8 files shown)


guidoagapito (approved):

specula/processing_objects/ext_source_pyramid.py
```diff
@@ -214,6 +214,14 @@ def __init__(self,
         # Add dedicated input for extended source coefficients
         self.inputs['ext_source_coeff'] = InputValue(type=BaseValue)
 
+    @staticmethod
```
guidoagapito: Using @classmethod allows us to extend parent dictionaries using super().input_names(). This eliminates the need to hardcode the parent class name, making the code much more resilient to changes in the hierarchy in the future.

Furthermore, class methods can be called directly on the class, just like static methods, without instantiation.

guidoagapito (changes requested):
See in line comments

specula/processing_objects/ext_source_pyramid.py
```diff
@@ -214,6 +214,14 @@ def __init__(self,
...
+    def input_names():
+        input_names = ModulatedPyramid.input_names()
+        input_names.update({
```
guidoagapito: We could replace the .copy() and .update() pattern with dictionary unpacking (e.g., {**super().input_names(), 'new_key': ...}).
This is a more concise, Pythonic approach that safely prevents accidental mutations of the parent class dictionary.

specula/processing_objects/ext_source_pyramid.py
```diff
@@ -214,6 +214,14 @@ def __init__(self,
...
+            'ext_source_coeff': (BaseValue, 'Extended source coefficients array of shape (n_points, 4)'
```
guidoagapito: We could replace the tuples with a NamedTuple for defining the inputs.

As we scale our doc-generation and validation tools, using named attributes rather than hardcoded tuple indices will make the code much easier to read and extend (e.g., if we want to add default_value or unit later).

specula/processing_objects/ext_source_pyramid.py
alfiopuglisi: done

specula/processing_objects/ext_source_pyramid.py
alfiopuglisi: It's also slower, because the current code has no copy, and in this case there is no concern of mutating the parent dictionary, I would keep the current code

specula/processing_objects/ext_source_pyramid.py
alfiopuglisi: I agree, in fact I was thinking the same, it is done now

specula/processing_objects/ext_source_pyramid.py
guidoagapito: Ok

guidoagapito (approved):

FabioRossiArcetri:
Have you tried fixing the failing test by putting a higher timeout in test_set_input_task_process, where it calls: obj.q.get(timeout=2) ?

alfiopuglisi:
> Have you tried fixing the failing test by putting a higher timeout in test_set_input_task_process, where it calls: obj.q.get(timeout=2) ?

Yes it has already included in a different PR, I will merge the main branch to get those changes and try again

FabioRossiArcetri (approved):

PR #578, opened 2026-04-01
EF input list in EF combinator

cedricplantet:
Added the possibility to give a list of EF instead of just two. The new code is retro-compatible.


test/test_electric_field.py
```diff
@@ -89,6 +89,34 @@ def test_ef_combinator(self, target_device_idx, xp):
...
+        ef_combinator_list.trigger()
+        ef_combinator_list.post_trigger()
+
+        out_ef_list = ef_combinator_list.outputs['out_ef']
+
+        # Assert that all 3 fields were properly combined
+        assert np.allclose(out_ef_list.A, ef1.A * ef2.A * ef3.A)
+        assert np.allclose(out_ef_list.phaseInNm, ef1.phaseInNm + ef2.phaseInNm + ef3.phaseInNm)
+        assert np.allclose(out_ef_list.S0, ef1.S0 + ef2.S0 + ef3.S0)
+
```
alfiopuglisi, line 119: I would add a test that sets both in_ef_list and in_ef1/in_ef2, and verifies that the setup method raises an exception

alfiopuglisi (approved):

guidoagapito (approved):

PR #579, opened 2026-04-02
Fixed synim==1.0.5

alfiopuglisi:
More recent versions make SPRINT tests fail


guidoagapito:
I reverted last commits in SynIM.

alfiopuglisi:
> I reverted last commits in SynIM.

But I think it is better to have a specific version in the requirements.txt anyway.
And not only for SynIM, but in principle for all dependencies.

guidoagapito (approved):

PR #581, opened 2026-04-02
Phase unwrapping update

bernadett92:
Updated the phase unwrapping algorithm in modal analysis to a least squares approach by solving the Poisson equation with the discrete cosine transform, because the previous algorithm sometimes failed for the physical propagation for bad seeing and large propagation distances.

### DIFF
--- a/specula/processing_objects/modal_analysis.py
+++ b/specula/processing_objects/modal_analysis.py
@@ -146,9 +146,7 @@ class ModalAnalysis(BaseProcessingObj):
 
 
     def unwrap_2d(self, p):
-        unwrapped_p = cpuArray(self.xp.copy(p))
-        unwrapped_p = self.unwrap_ls(unwrapped_p)
-
+        unwrapped_p = self.unwrap_ls(cpuArray(p))
         return self.to_xp(unwrapped_p)
 
     def setup(self):
--- a/specula/processing_objects/modal_analysis.py
+++ b/specula/processing_objects/modal_analysis.py
@@ -12,7 +12,11 @@ from scipy.fftpack import idct, dct
 import numpy as np
 
 class ModalAnalysis(BaseProcessingObj):
-
+    """
+    Modal analysis processing object. Decomposition of an
+    input ElectricField into modes defined
+    by an influence function (e.g. Zernike polynomials).
+    """
     def __init__(self,
                 ifunc: IFunc=None,
                 ifunc_inv: IFuncInv=None,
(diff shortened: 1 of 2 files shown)


specula/processing_objects/modal_analysis.py
```diff
@@ -101,9 +102,53 @@ def prepare_trigger(self, t):
...
+        denom[0, 0] = 1.0
+        dct_phi = dct_rho / denom
+        dct_phi[0, 0] = 0.0 # avoid division by zero
+
+        # Inverse 2D DCT
+        return idct(idct(dct_phi, axis=0, norm='ortho'), axis=1, norm='ortho')
+
+
     def unwrap_2d(self, p):
         unwrapped_p = cpuArray(self.xp.copy(p))
```
alfiopuglisi: I don't think this self.xp.copy is needed, just cpuArray(p)

test/test_modal_analysis_unwrapping.py
```diff
@@ -38,7 +38,7 @@ def test_modal_analysis_unwrapping(self, target_device_idx, xp):
                                      target_device_idx=target_device_idx)
 
         # Physical and geometrical propagation to source
-        uplink_source = Source(polar_coordinates=[0.0, 0.0], magnitude=0, height=300,
+        uplink_source = Source(polar_coordinates=[0.0, 0.0], magnitude=0, height=5000,
```
alfiopuglisi, line 99: Is this height=5000 test case one of those where the previous approach was failing? It would be nice to have a test where only this new algorithm succeeds

test/test_modal_analysis_unwrapping.py
```diff
@@ -75,4 +75,5 @@ def test_modal_analysis_unwrapping(self, target_device_idx, xp):
 
         modes_phys = cpuArray(modal_analsis_phys.outputs['out_modes'].value)
         modes_geom = cpuArray(modal_analsis_geom.outputs['out_modes'].value)
-        np.testing.assert_allclose(modes_phys, modes_geom, rtol=3)
+        rel_error = np.mean(abs((modes_phys - modes_geom))/abs(modes_geom))
+        np.testing.assert_array_less(rel_error, 0.2)
```
alfiopuglisi, line 137: Is this a 20% tolerance? It looks a bit high, unless this is a feature of the algorithm...

test/test_modal_analysis_unwrapping.py
bernadett92: Yes, this is a test case were the previous one was failing. I also added another test case were I wrap the phase and compare the new unwrapping with the skimage one.

test/test_modal_analysis_unwrapping.py
bernadett92: This relatively high error is because I compare physical and geometrical propagation, which is expected to be different. I added another test case were I just compare the phase unwrapping.

alfiopuglisi (approved):

guidoagapito (approved):

PR #582, opened 2026-04-02
Updated EF Combinator

alfiopuglisi:
Simplified logic.
Added checking of matching pixel pitch as well
Added various tests

### DIFF
--- a/specula/processing_objects/electric_field_combinator.py
+++ b/specula/processing_objects/electric_field_combinator.py
@@ -48,66 +48,46 @@ class ElectricFieldCombinator(BaseProcessingObj):
                 "or BOTH 'in_ef1' and 'in_ef2' to be connected."
             )
 
-        # Priority check: if a list is provided, use it to configure output shape
         if has_list:
             first_ef = in_ef_list[0]
-            
             # Verify that all provided fields have matching shapes
             for i, ef in enumerate(in_ef_list[1:]):
                 if first_ef.A.shape != ef.A.shape:
                     raise ValueError(f"Input electric field list index {i+1} shape {ef.A.shape} does not match index 0 shape {first_ef.A.shape}")
-
-            self._out_ef.resize(
-                dimx=first_ef.A.shape[0],
-                dimy=first_ef.A.shape[1],
-                pitch=first_ef.pixel_pitch,
-            )
-            # Return early to bypass the legacy logic
-            return
-
-        # Legacy execution: Pair configuration
-        if in_ef1.A.shape != in_ef2.A.shape:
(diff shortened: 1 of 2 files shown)


specula/processing_objects/electric_field_combinator.py
```diff
@@ -48,66 +48,46 @@ def setup(self):
...
-        # Multiply amplitudes
-        self._out_ef.A[:] = in_ef1.A * in_ef2.A
-
-        # Combine S0 values
-        self._out_ef.S0 = in_ef1.S0 + in_ef2.S0
-
-        # Set the generation time to the current time
+        # Accumulate values from the rest of the list
+        for in_ef in in_ef_list[1:]:
+            self._out_ef.phaseInNm += in_ef.phaseInNm
```
guidoagapito, line 89: I think we need [:] here and at the following line, or am I wrong?

cedricplantet: I don't really see why, but I still have a lot to learn in Python. In my case I put [:] just to be consistent with the rest of the code.

guidoagapito: [:] is required to update the content of the array in-place without creating a new copy.

cedricplantet: Ok, good to know! Then it's missing indeed.

alfiopuglisi: No, it's more subtle than that: in Python, an assignment replaces the assigned value with a reference to the new value, so:

self.out_ef.phaseInNm = other

will replace phaseInNm with "other", and it will end up as an array with a different memory address from before. You can check with id(other) and id (self.out_ef.phaseInNm): they will be different before, and the same afterwards. Thd old array will be automatically discarded. While instead

self.out_ef.phaseInNm[:] = other

will replace the elements of phaseInNm, rather than the array itself. In other words it will overwrite the contents of phaseInNm, leaving it in place, which is exactly what we want.

However, this is only needed for assignment. An operator like "+=" is changing the contents of phaseInNm in place, overwriting the new values, so there is no need for [:] (it still works, but it's redundant).

You can see for example the "apply_noise" method of ccd.py where the same distinction is used for the various overwriting/replacements.

In summary, we can keep the code as it is.

alfiopuglisi: As a consequence, in Python these two statements:

a = a +b
a += b

are not equivalent. The first one is creating a third array and discarding the old "a", the second is updating "a" in place without reallocating.

cedricplantet: Ok, thanks for the explanation!

cedricplantet (approved):

guidoagapito (approved):

PR #590, opened 2026-04-08
(GA) compute radial profile, FWHM and EE in PSF

guidoagapito:


specula/processing_objects/psf.py
```diff
@@ -136,26 +217,42 @@ def post_trigger(self):
...
+            variance = self._sum_psf_squared / self.count - self.int_psf.value ** 2
+            self.std_psf.value = self.xp.sqrt(self.xp.maximum(variance, 0))
+            if self.compute_profile_metrics:
+                self._store_profile_metrics(
+                    self.int_psf.value,
+                    self.psf_profile,
+                    self.psf_fwhm,
+                    self.encircled_energy,
+                    self.encircled_energy_at_radius,
+                )
```
alfiopuglisi, line 271: This last store call suddenly replaces the frame-by-frame metrics with different ones based on the integrated PSF. Is this the intended behaviour?

specula/processing_objects/psf.py
```diff
@@ -167,7 +264,32 @@ def get_psf_profile(self, psf_std:bool=False, show:bool=False):
                 plt.xlabel(r'$\lambda/D$')
                 plt.ylabel('PSF contrast')
                 plt.show()
-            except:
+            except Exception:
```
alfiopuglisi: Should be except ImportError

specula/lib/radial_profile.py
```diff
@@ -1,39 +1,207 @@
 import numpy as np
 
-def computeRadialProfile(image, centerInPxY=None, centerInPxX=None, xp=np, dtype=np.float64):
+
+def computeRadialProfile(image, centerInPxY=None, centerInPxX=None,
```
alfiopuglisi: These function names have been basically copied from arte, but I think we should change them to the standard pep8 camel case: compute_radial_profile, etc

guidoagapito: Done

specula/processing_objects/psf.py
guidoagapito: These methods have been removed.

specula/processing_objects/psf.py
guidoagapito: No, thanks. New outputs have been added.

specula/processing_objects/psf_coronagraph.py
```diff
@@ -67,10 +79,34 @@ def __init__(self,
...
+        self.outputs['out_coronagraph_psf_fwhm'] = self.coronagraph_psf_fwhm
+        self.outputs['out_coronagraph_encircled_energy'] = self.coronagraph_encircled_energy
+        self.outputs['out_coronagraph_encircled_energy_at_radius'] = self.coronagraph_encircled_energy_at_radius
+        self.outputs['out_int_coronagraph_psf_profile'] = self.int_coronagraph_psf_profile
+        self.outputs['out_int_coronagraph_psf_fwhm'] = self.int_coronagraph_psf_fwhm
+        self.outputs['out_int_coronagraph_encircled_energy'] = self.int_coronagraph_encircled_energy
+        self.outputs['out_int_coronagraph_encircled_energy_at_radius'] = self.int_coronagraph_encircled_energy_at_radius
 
         # Reference complex amplitude for perfect coronagraph
         self.ref_complex_amplitude = None
```
Menessao, line 98: Not strictly related to the PR, but I have also noticed a bug in my previous commit of the psf_coronagraph

I think the current lines 123-124 (should be around 157-158 after your modifications):
```
else: # perfect coronagraph formula (Cavarroc et al. 2006, Eq. 1)
     [...]
    ec = self.xp.exp(-var_phase, self.dtype)
    coherent_core = self.xp.sqrt(ec) * self.xp.exp(1j * mean_phase, self.complex_dtype) * amp
    [...]
```  
should have the dtype= argument explicit to avoid issues, hence becoming:

```
else: # perfect coronagraph formula (Cavarroc et al. 2006, Eq. 1)
     [...]
    ec = self.xp.exp(-var_phase, dtype=self.dtype)
    coherent_core = self.xp.sqrt(ec) * self.xp.exp(1j * mean_phase, dtype=self.complex_dtype) * amp
    [...]
```

specula/processing_objects/psf_coronagraph.py
```diff
@@ -67,10 +79,34 @@ def __init__(self,
...
+        self.coronagraph_encircled_energy_at_radius = BaseValue(target_device_idx=self.target_device_idx,
+                                                                precision=precision)
+        self.int_coronagraph_psf_profile = BaseValue(target_device_idx=self.target_device_idx,
+                                                     precision=precision)
+        self.int_coronagraph_psf_fwhm = BaseValue(target_device_idx=self.target_device_idx,
+                                                  precision=precision)
+        self.int_coronagraph_encircled_energy = BaseValue(target_device_idx=self.target_device_idx,
+                                                          precision=precision)
+        self.int_coronagraph_encircled_energy_at_radius = BaseValue(target_device_idx=self.target_device_idx,
+                                                                    precision=precision)
```
Menessao: I might be wrong about this, but I am not sure whether EE and FWHM are of much interest for a perfect coronagraph. I have to aknowledge, though, that is only an issue if their computation is time expensive.

Additionaly, one performance parameter that is often used is the radial profile of the STD of the coronagraphic PSF, which is currently not computed, unless I have missed it.

The profiles of coronagraphic PSFs (both int and STD) would also be very useful in the abstract_coronagraph class

specula/processing_objects/psf_coronagraph.py
guidoagapito: Done

Menessao (commented):
I think these are very good additions to the current code, I could add something similar to the abstract_coronagraph class unless you want to do it in this PR

specula/processing_objects/psf_coronagraph.py
guidoagapito: For abstract_coronagraph I think we should open a different pull request.

I agree on FWHM, EE and radial profile of STD.

alfiopuglisi (approved):

PR #591, opened 2026-04-08
Bug fixes in LIFT

cedricplantet:
Corrected several calculations linked to the PSF sampling and centering.
Changed the formula of radians_per_pixel (wrong by a factor 2).
Fixed a bug in set_modalbasis: the modes were converted to int if the mask in the input ifunc was an int array.
Added a "debug" keyword to display the analyzed PSF and the reconstructed one at each iteration.


specula/processing_objects/lift.py
```diff
@@ -42,6 +43,7 @@ def __init__(self,
                  n_iter: int=20,
                  fft_res: int=2,
                  fix: bool=False,
+                 debug: bool=False,
```
guidoagapito: We have never used a debug argument in a processing object, but we have processing object with hardcoded keywords in the methods.

@alfiopuglisi what do you think?

specula/processing_objects/lift.py
```diff
@@ -233,7 +235,7 @@ def set_modalbase(self, modalbase, mask2d, diameter):
         )
 
         valid_idx = np.nonzero(mask2d)
-        f = np.zeros_like(mask2d)
+        f = np.zeros_like(mask2d.astype(cpu_dtype))
```
guidoagapito, line 236: It would be nice to have a test for this bugfix.

FabioRossiArcetri (changes requested):
I would remove the plot, as a general rule for the processing objects, and replace debug with verbose.

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #594, opened 2026-04-10
Fixed bug in phasescreen error path

alfiopuglisi:


alfiopuglisi:
BTW, shouldn't this routine raise an exception instead of returning None if the calculation fails? @guidoagapito

guidoagapito:
> BTW, shouldn't this routine raise an exception instead of returning None if the calculation fails? @guidoagapito

Yes, you are right, it should raise an exception.

guidoagapito (approved):

specula/lib/calc_phasescreen.py
```diff
@@ -89,11 +87,10 @@ def calc_phasescreen(L0, dimension, pixel_pitch, xp, precision, seed=0, verbose=
     # Check for non-finite elements and handle them
```
alfiopuglisi: @guidoagapito isn't this last check redundant? "phasescreen" is filled using re_gauss, which has already been checked and should contain no non-finite elements

guidoagapito: yes, you are right.

FabioRossiArcetri (approved):

guidoagapito (approved):

PR #603, opened 2026-04-13
Update angular spectrum propagation

bernadett92:
Updated angular spectrum propagation algorithm to be numerically more stable. No automatic reduction of propagation distance like before.


specula/processing_objects/atmo_propagation.py
```diff
@@ -263,6 +248,11 @@ def trigger_code(self):
 
         for source_name, source in self.source_dict.items():
 
+            # reset field
+            if self.doFresnel:
+                self.ef_fresnel[:] *= 0
+                self.ef_fresnel[:] += 1
+
```
alfiopuglisi, line 289: Is this equivalent to:

self.ef_fresnel[:] =1

? it should be faster

specula/processing_objects/atmo_propagation.py
```diff
@@ -279,15 +269,30 @@ def trigger_code(self):
...
                                (layer.size[1] - self.pixel_pupil_size) // 2]
-                    output_ef.product(layer, subrect=topleft)
+                    x2 = topleft[0] + output_ef.size[0]
+                    y2 = topleft[1] + output_ef.size[1]
+                    self.phi[:] = layer.field[1, topleft[0]: x2, topleft[1]: y2]
+                    self.amp[:] = layer.field[0, topleft[0]: x2, topleft[1]: y2]
                 else:
-                    output_ef.A *= interpolator.interpolate(layer.A)
-                    output_ef.phaseInNm += interpolator.interpolate(layer.phaseInNm)
+                    self.phi[:] = (interpolator.interpolate(layer.phaseInNm))
```
alfiopuglisi: Too many parenthesis, can remove the outer one

specula/processing_objects/atmo_propagation.py
bernadett92: Yes, it is. I changed it.

specula/processing_objects/atmo_propagation.py
```diff
@@ -157,46 +153,29 @@ def __init__(self,
...
-        H_AS = self.xp.exp(-1j * np.pi * distanceInM * self.wavelengthInNm * 1e-9 * fsq)
-
-        # Apply bandlimit filter
-        if self.band_limit_factor < 1.0:
-            W = ((fx / f_limit) ** 2 + (fy * self.wavelengthInNm * 1e-9) ** 2 <= 1) * (
-                    (fy / f_limit) ** 2 + (fx * self.wavelengthInNm * 1e-9) ** 2 <= 1)
-            H_AS *= W
-
-        return H_AS
+    # Scalable Angular Spectrum Propagation Method
```
guidoagapito: I think it will be useful to add some more details in the doc string about this method and the advantages with respect to other methods.

test/test_physical_propagation.py
```diff
@@ -1,161 +1,180 @@
...
-            for obj in objlist:
-                obj.post_trigger()
-
-        uplink_phase1 = prop_up1.outputs['out_uplink_source_ef'].phaseInNm
-        uplink_phase2 = prop_up2.outputs['out_uplink_source_ef'].phaseInNm
-
-        rms = xp.sqrt(xp.mean((uplink_phase1 / np.max(uplink_phase1) - uplink_phase2 / np.max(uplink_phase2)) ** 2))
-        print(rms)
-        self.assertTrue(rms < 0.1)
+import specula
```
guidoagapito, line 3: Do we have a test where the old method was not numerically stable?

We could also compare the two methods in one test.

specula/processing_objects/atmo_propagation.py
```diff
@@ -279,15 +268,30 @@ def trigger_code(self):
                 if interpolator is None:
                     topleft = [(layer.size[0] - self.pixel_pupil_size) // 2, \
                                (layer.size[1] - self.pixel_pupil_size) // 2]
-                    output_ef.product(layer, subrect=topleft)
+                    x2 = topleft[0] + output_ef.size[0]
+                    y2 = topleft[1] + output_ef.size[1]
+                    self.phi[:] = layer.field[1, topleft[0]: x2, topleft[1]: y2]
```
guidoagapito: You could use an electric field instead of amp and phi, but we would need a new method to extract a subrect.

specula/processing_objects/atmo_propagation.py
```diff
@@ -279,15 +268,30 @@ def trigger_code(self):
...
                 else:
-                    output_ef.A *= interpolator.interpolate(layer.A)
-                    output_ef.phaseInNm += interpolator.interpolate(layer.phaseInNm)
+                    self.phi[:] = interpolator.interpolate(layer.phaseInNm)
+                    self.amp[:] = interpolator.interpolate(layer.A)
+
+                if self.doFresnel:
+                    if self.upwards:
+                        self.phi *= -1
+                    self.ef_fresnel *= self.amp * self.xp.exp(1j * self.phi / self.wavelengthInNm * 2 * self.xp.pi, dtype=self.complex_dtype)
```
guidoagapito: If you have and electic field instead of phi and amp you could use the ef_at_lambda method.

bernadett92: I did not use an electric field, because then the angle function is applied for every propagation. For physical propagation I need the electric field in the form A*exp(1j*phi) and to set the electric field phase I need to apply the angle function. I wanted to keep the form A*exp(1j*phi) during the propagation and set the electric field only once at the end.

guidoagapito: What I meant is to use an explicit electric field array (let's call it temp_ef or ef), instead of keeping amplitude (amp) and phase (phi) as separate arrays. This electric field would directly represent the complex field at each step, combining amplitude and phase:

`ef = amp * exp(1j * phi / wavelengthInNm * 2 * pi)`

This approach has several advantages:

The electric field (ef) contains all the information (both amplitude and phase) in a single array, making propagation and transformations more natural and less error-prone.
You can define methods on this object/array for propagation, phase shifts, or amplitude changes, making the code more modular and readable.
When you need to extract amplitude or phase, you can always get them as abs(ef) and angle(ef), but most calculations (like propagation) work directly on ef.
So, instead of updating amp and phi separately and then combining them only at the end, you update ef at each step, applying the relevant transformations (multiplications, propagators, etc.) directly.

bernadett92: For doFresnel=True the electric field is already kept in this form to avoid numerical errors, but if doFresnel=False I don't have the wavelengthInNm (this is just an optional parameter). This is why I introduced phi and amp.

guidoagapito: Ok, but you do not need the wavelengthInNm to define an electric field object nor to manipulate it.
wavelengthInNm is required only when phi_at_lambda or ef_at_lambda are called.

What I mean is more or less this:

```
            for li, layer in enumerate(layer_list):
                if not self.mergeLayersContrib:
                    output_ef = output_ef_list[li]
                    output_ef.reset()

                interpolator = self.interpolators[source][layer]
                if interpolator is None:
                    topleft = [(layer.size[0] - self.pixel_pupil_size) // 2, \
                               (layer.size[1] - self.pixel_pupil_size) // 2]
                    x2 = topleft[0] + output_ef.size[0]
                    y2 = topleft[1] + output_ef.size[1]
                    self.ef_temp.A[:] = layer.A[topleft[0]: x2, topleft[1]: y2]
                    self.ef_temp.phaseInNm[:] = self.prop_sign * layer.phaseInNm[topleft[0]: x2, topleft[1]: y2]
                else:
                    self.ef_temp.A[:] = interpolator.interpolate(layer.A)
                    self.ef_temp.phaseInNm[:] = self.prop_sign * interpolator.interpolate(layer.phaseInNm)

                if self.doFresnel:
                    self.ef_fresnel *= self.ef_temp.ef_at_lambda(self.wavelengthInNm)
                    if self.propagators[li] is not None:
                        self.angular_spectrum_propagation(self.ef_fresnel, self.propagators[li])
                else:
                    output_ef.A *= ef_temp.A
                    output_ef.phaseInNm += ef_temp.phaseInNm

            if self.doFresnel:
                output_ef.phaseInNm[:] = self.prop_sign * self.xp.angle(self.ef_fresnel) * self.wavelengthInNm / (
                            2 * self.xp.pi)
                output_ef.A[:] = abs(self.ef_fresnel)
```

test/test_physical_propagation.py
bernadett92: The propagator itself is the same. I added a test which compares the previous propagator implementation with the new one. I removed the automatic reduction of the propagation distance. The numerically more stable was meant for the angle function, which was previously applied in the loop and introduced some errors.

specula/processing_objects/atmo_propagation.py
bernadett92: I implemented it in this way.

guidoagapito:
Thank you for your efforts and for following all the advice.

guidoagapito (approved):

PR #609, opened 2026-04-15
(GA) input and output lists in processing object summary

guidoagapito:


docs/scripts/generate_objects_summary.py
```diff
@@ -4,21 +4,227 @@
 from pathlib import Path
 
 
-def get_class_short_doc(filepath):
-    """Return a dict {classname: short_docstring} for all classes in a file."""
+def expr_to_string(node):
```
alfiopuglisi: This is a rather complicated code that uses the Abstract Source Tree module, that is, the internal Python representation of sources.
It should be possible to do all this in a much simpler way, by iterating on the classes like we do in specula_testlib (maybe extracting the same functions in a common lib) and just instantiating them and using the classmethods to get input and output names

guidoagapito: Ok, you are right, I'm working on it.

alfiopuglisi (approved):

PR #616, opened 2026-04-19
Modal analysis on GPU

alfiopuglisi:


test/test_modal_analysis_unwrapping.py
```diff
@@ -20,8 +20,6 @@
 
 import numpy as np
 
-@unittest.skipIf(os.getenv('CI') == 'true',
```
guidoagapito, line 23: Apparently we need this.

alfiopuglisi: It's a threading problem when using scipy DCT, I am trying some workarounds

alfiopuglisi: In the end none of the workarounds worked. I disabled this test just for the few problematic configurations (ubuntu with python versions from 3.11 to 3.13), it is enabled on all others. Anyway it only happens on Github and not on a linux workstation.

guidoagapito (approved):

PR #596, opened 2026-04-10
(GA) multirate modalrec and _dict_object in simul with tests

guidoagapito:
Please note that:
- I added _list_ref and _list_objects cases in simul.py
- I updated MultirateComplementaryFilter

PR #617, opened 2026-04-19
List dict object

alfiopuglisi:
Removed complex typing code, using typing.get_args instead


specula/simul.py
```diff
@@ -21,49 +22,6 @@
 
 
 Output = namedtuple('Output', 'obj_name output_key delay ref input_name')
-UNION_ORIGINS = (typing.Union,) + ((types.UnionType,) if hasattr(types, 'UnionType') else ())
```
guidoagapito, line 26: I remember that this part was added to provide support for python 3.8.

I took some time to find a solution.
You can see the steps in the original pull request: https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/596

alfiopuglisi: I found a simpler solution using get_origin and get_args from the typing module, it is now in the resolve_type function in specula/lib/utils.py in this PR.

guidoagapito (approved):

PR #627, opened 2026-04-27
(GA) PupilstopController with test

guidoagapito:


specula/processing_objects/pupilstop_controller.py
```diff
@@ -0,0 +1,171 @@
...
+                magnification=float(self._pupilstop.magnification),
+            )
+            self._ef_interpolator.interpolate()
+            mask = self._ef_interpolator.interpolated_ef().A
+
+            mask = self.xp.asarray(mask, dtype=self._pupilstop.dtype)
+            if self.threshold_mask:
+                mask = (mask >= self.mask_threshold).astype(self._pupilstop.dtype)
+
+            self._pupilstop.A[:] = mask
```
alfiopuglisi: This is overwriting the input pupilstop, which is not a good idea (it could be used somewhere else in the same simulation).
It would be better to define a basic output layer in __init__,  setup an interpolator if some of the inputs are valid, and call the interpolator if needed in trigger

guidoagapito: Done

specula/processing_objects/pupilstop_controller.py
```diff
@@ -0,0 +1,172 @@
...
+        target_device_idx : int, optional
+            Target device index for computation (CPU/GPU). Default is None
+            (uses global setting).
+        precision : int, optional
+            Precision for computation (0 for double, 1 for single). Default is None
+            (uses global setting).  
+        """
+        super().__init__(target_device_idx=target_device_idx, precision=precision)
+
+        if pupilstop.target_device_idx != self.target_device_idx:
```
alfiopuglisi: Why the same? Is it not possible to use to_xp() ?

specula/processing_objects/pupilstop_controller.py
```diff
@@ -0,0 +1,172 @@
...
+            pixel_pitch=pupilstop.pixel_pitch,
+            height=0,
+            shiftXYinPixel=cpuArray(pupilstop.shiftXYinPixel).astype(float),
+            rotInDeg=float(pupilstop.rotInDeg),
+            magnification=float(pupilstop.magnification),
+            target_device_idx=self.target_device_idx,
+            precision=self.precision,
+        )
+        self._out_layer.A[:] = pupilstop.A
+        self._out_layer.generation_time = 0
```
alfiopuglisi: I think this setting of generation_time is redundant, it will be done in post_trigger in any case

specula/processing_objects/pupilstop_controller.py
guidoagapito: Done

specula/processing_objects/pupilstop_controller.py
guidoagapito: Removed

specula/processing_objects/pupilstop_controller.py
```diff
@@ -0,0 +1,179 @@
...
+            dimy=pupilstop.size[0],
+            pixel_pitch=pupilstop.pixel_pitch,
+            height=0,
+            shiftXYinPixel=in_shift_xy,
+            rotInDeg=float(pupilstop.rotInDeg),
+            magnification=float(pupilstop.magnification),
+            target_device_idx=self.target_device_idx,
+            precision=self.precision,
+        )
+        self._in_layer.A[:] = self.to_xp(pupilstop.A, dtype=self._in_layer.dtype, force_copy=True)
```
alfiopuglisi: dtype and force_copy should be unnecessary: the assignment to [:] should enforce both

specula/processing_objects/pupilstop_controller.py
```diff
@@ -0,0 +1,179 @@
...
+            dimy=pupilstop.size[0],
+            pixel_pitch=pupilstop.pixel_pitch,
+            height=0,
+            shiftXYinPixel=in_shift_xy,
+            rotInDeg=float(pupilstop.rotInDeg),
+            magnification=float(pupilstop.magnification),
+            target_device_idx=self.target_device_idx,
+            precision=self.precision,
+        )
+        self._out_layer.A[:] = self.to_xp(pupilstop.A, dtype=self._out_layer.dtype, force_copy=True)
```
alfiopuglisi: Same comment as above

specula/processing_objects/pupilstop_controller.py
```diff
@@ -0,0 +1,179 @@
...
+                magnification=float(self._out_layer.magnification),
+            )
+            self._ef_interpolator.interpolate()
+            mask = self._ef_interpolator.interpolated_ef().A
+
+            mask = self.xp.asarray(mask, dtype=self._out_layer.dtype)
+            if self.threshold_mask:
+                mask = (mask >= self.mask_threshold).astype(self._out_layer.dtype)
+
+            self._out_layer.A[:] = mask
```
alfiopuglisi, line 173: the [:] here is already performing dtype conversion, the two dtype conversions before should be unnecessary

specula/processing_objects/pupilstop_controller.py
guidoagapito: Removed

specula/processing_objects/pupilstop_controller.py
guidoagapito: Removed

specula/processing_objects/pupilstop_controller.py
guidoagapito: Removed

alfiopuglisi (approved):

PR #633, opened 2026-04-27
Several small simplifications

alfiopuglisi:


specula/processing_objects/iir_filter.py
guidoagapito, line 36: I think this (simul_params) is not used, but was kept for compatibility with yml files.
We could remove it.
This is valid for ssr_filter too and for: `self.time_step = simul_params.time_step` in base_filter.

guidoagapito: I add: if we remove this we need to update yml files and documentation / tutorials.

alfiopuglisi: Ok, it's in fact a rather big change, it could be discussed separately, for now I have reverted this part.

guidoagapito (approved):

FabioRossiArcetri (approved):

PR #639, opened 2026-04-28
Scalar values

alfiopuglisi:
Add specific data objects for scalar values: int, float and string.


guidoagapito (approved):

specula/processing_objects/specula_input.py
```diff
@@ -60,10 +60,12 @@ def trigger_code(self):
             while True:
                 name, value = self.q.get(block=False)
                 try:
-                    # Replace value directly instead of using set_value
-                    # so that we can accept any valid type instead of just numeric ones
-                    self.outputs[name].value = value
-                    self.outputs[name].generation_time = self.current_time
+                    cast_func = self.outputs[name].type
```
guidoagapito, line 74: Maybe self.outputs[name].type --> self.outputs[name].value.type ?

guidoagapito (changes requested):

specula/processing_objects/specula_input.py
alfiopuglisi: It uses ".type" as an explicit member instead of using __ type__ of the value,  since the latter is not defined for some built-ins (like int)

guidoagapito (approved):

FabioRossiArcetri (approved):

alfiopuglisi:
Before reintegrating, there is a detail to fix: now the ":" syntax is used for the loop delay (dm.commands:-1) and for the type (gain:int). It shouldn't be a problem because the delay is for inputs only, and the type for outputs only, but the parsing should enforce this

PR #642, opened 2026-04-30
(GA) Fix chromatic shifts in setup_interpolators

guidoagapito:


alfiopuglisi:
It seems that AtmoPropagation.compute_chromatic_shifts() is modifying the input Source objects, storing results there.
I don't think it's a good idea: the results should be stored instead in the AtmoPropagation instance itself, without touching the input data.

FabioRossiArcetri (approved):

guidoagapito:
> It seems that AtmoPropagation.compute_chromatic_shifts() is modifying the input Source objects, storing results there. I don't think it's a good idea: the results should be stored instead in the AtmoPropagation instance itself, without touching the input data.

Done.

specula/processing_objects/atmo_propagation.py
```diff
@@ -364,7 +364,6 @@ def compute_chromatic_shifts(self, source, atmo_layer_list):
         If enable_chromatic_effect is False or the two wavelengths are identical,
         all shifts are zero.
         """
-        source.chromatic_shifts_m = {}
```
alfiopuglisi, line 367: Should we remove source.chromatic_shifts_m from the Source init function as well?

guidoagapito: Done

alfiopuglisi (approved):

PR #646, opened 2026-05-04
Fix radial profile metrics for PSF and coro PSF

Menessao:
Corrected errors returning types instead of values

Re-added verbose option to PSF to optinonally avoid printing multiple lines (PSF automatically prints SR, coro PSF printed suppression) each iteration

Fixed normalization issue in coro profile: the issue was due to the fact that the radial profile was always normalized to its maximum value. 
This is probably correct for the standard PSF (and the correct scaling can be easily recovered from the SR) but, when it comes to the coronagraphic PSF, the suppression wrt to the PSF peak not only should (hopefully) be much lower than 1, but is also important information


specula/processing_objects/psf.py
```diff
@@ -63,6 +63,7 @@ def __init__(self,
                  ee_radius_in_lambda_d=None,
                  target_device_idx: int = None,
                  precision: int = None,
+                 verbose:bool = True,
```
guidoagapito: We no longer use the keyword “verbose”, see https://github.com/ArcetriAdaptiveOptics/SPECULA/pull/575

specula/processing_objects/psf_coronagraph.py
```diff
@@ -58,6 +58,7 @@ def __init__(self,
                  ee_radius_in_lambda_d=None,
                  target_device_idx: int = None,
                  precision: int = None,
+                 verbose: bool = False,
```
guidoagapito: see other comment.

specula/lib/radial_profile.py
guidoagapito, line 96: I see you removed dtype elsewhere, why not here?

Nevertheless, I think that xp.array(np.nan, dtype=dtype) (and xp.array(0, dtype=dtype), xp.array(2.0 * half_radius, dtype=dtype)) is safer than dtype(np.nan) for array backends other than NumPy, but still raises for integer dtypes — which are not meaningful for a FWHM computation anyway.

guidoagapito: if p2 == p1:

Menessao: Added in the last commit. I am not sure whether the fact that we are casting scalar values to XP.arrays causes issues elsewhere, but it probably doesn't, and is an easy way to have the correct dtype

specula/lib/radial_profile.py
```diff
@@ -79,32 +79,32 @@ def compute_fwhm_from_profile(profile, radial_distance=None, xp=np, dtype=np.flo
...
     p2 = profile[idx]
 
-    if float(p2 - p1) == 0.0:
+    if p1 == p2:
         half_radius = r1
     else:
         # Linear interpolation to find the radius at half maximum
         half_radius = r1 + (half_maximum - p1) * (r2 - r1) / (p2 - p1)
-    return dtype(2.0) * dtype(half_radius)
+    return xp.array(2.0 * half_radius,dtype=dtype)
```
guidoagapito: I'm sorry, I suggested this change, but it isn't necessary here: 2.0 * half_radius is sufficient.

guidoagapito (approved):

alfiopuglisi (approved):

PR #651, opened 2026-05-12
(GA) removed ifunc cuts in ModalAnalysis

guidoagapito:
ifunc.cut(nmodes=nmodes) modifies the object in place, so reusing the same instance elsewhere would also reflect the truncation.
This pull request changes the behaviour to prevent this from happening.

### DIFF
--- a/specula/processing_objects/modal_analysis.py
+++ b/specula/processing_objects/modal_analysis.py
@@ -56,7 +56,7 @@ class ModalAnalysis(BaseProcessingObj):
             self.phase2modes = ifunc.inverse()
         elif ifunc is None and ifunc_inv is not None:
             # Use ifunc_inv directly, don't attempt to call inverse() on None
-            if nmodes is not None and nmodes != ifunc_inv.size[0]:
+            if nmodes != ifunc_inv.nmodes():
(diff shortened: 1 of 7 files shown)


specula/processing_objects/modal_analysis.py
```diff
@@ -57,13 +57,14 @@ def __init__(self,
         elif ifunc is None and ifunc_inv is not None:
             # Use ifunc_inv directly, don't attempt to call inverse() on None
             if nmodes is not None and nmodes != ifunc_inv.size[0]:
-                ifunc_inv.cut(nmodes=nmodes)
+                ifunc_inv = IFuncInv(ifunc_inv.ifunc_inv[:, :nmodes],
```
alfiopuglisi, line 60: It looks like it should be ifunc_inv.size[1].
Even better would be to add nmodes() and npoints() to IFunc and IFuncInv to get the dimensions

alfiopuglisi: Also there is no need to check for None, it will result in the whole array being used, which is the wanted result,.

guidoagapito: Done

alfiopuglisi (approved):

PR #652, opened 2026-05-13
(GA) MmsePistonUnwrapper, IirFilter in_ost input and tests

guidoagapito:
This pull request introduces a new processing object for MMSE piston unwrapping, enhances the IIR filter with an optional state update input, and improves test coverage for these features. The main themes are the addition of new functionality for piston unwrapping, generalization and extensibility of the IIR filter, and improved test coverage and robustness.


specula/processing_objects/mmse_piston_unwrapper.py
```diff
@@ -0,0 +1,100 @@
...
+
+        # Expect recmat_list[0] to be the MMSE reconstructor (5xN_modes) and
+        #        recmat_list[1] to be the Interaction matrix (N_modes x 5)
+        self.logger.info("Loading MMSE reconstructor and Interaction matrices from recmat_list...")
+        recmat_obj = recmat_list[0]
+        intmat_obj = recmat_list[1]
+
+        self.recmat = self.to_xp(recmat_obj.recmat, dtype=self.dtype) # Shape: (5, N_modes)
+        self.intmat = self.to_xp(intmat_obj.recmat, dtype=self.dtype) # Shape: (N_modes, 5)
+
```
alfiopuglisi, line 51: Is it reasonable to check that the two shapes match?

specula/processing_objects/mmse_piston_unwrapper.py
```diff
@@ -0,0 +1,100 @@
...
+        in_comm = self.local_inputs['in_comm'].value
+
+        out_comm_val = in_comm.copy()
+        out_ost_val = self.xp.zeros_like(in_comm)
+
+        do_check = False
+        if self.current_time >= self.start_time_t:
+            if self.interval_time_t <= 0:
+                do_check = True
+            elif (self.current_time - self.last_correction_time) >= self.interval_time_t - 1e-6:
```
alfiopuglisi: Is this check with "epsilon "(1e-6) really needed? all the variables in this line should be already converted to t, so they should be integers and comparisons should be exact. Is that not the case?

specula/processing_objects/mmse_piston_unwrapper.py
```diff
@@ -0,0 +1,100 @@
...
+        self.gain = gain
+
+        # SPECULA discrete time management
+        self.start_time_t = self.seconds_to_t(start_time)
+        self.interval_time_t = self.seconds_to_t(interval_time)
+        self.last_correction_time = -1
+
+        self.inputs['in_comm'] = InputValue(type=BaseValue)
+
+        self.outputs['out_comm'] = BaseValue(target_device_idx=self.target_device_idx,
```
alfiopuglisi, line 38: out_comm and out_ost are not allocated until trigger, which can give problems to downstream objects that check them during setup().,
They can be allocated in the setup() method here as well, which would also allow checking that the size agree with recmat and intmat

specula/processing_objects/mmse_piston_unwrapper.py
```diff
@@ -0,0 +1,100 @@
...
+        return {
+            'out_comm': OutputDesc(BaseValue, 'Cleaned command vector for the DM'),
+            'out_ost': OutputDesc(BaseValue, 'State correction vector for the IirFilter')
+        }
+
+    def trigger_code(self):
+        in_comm = self.local_inputs['in_comm'].value
+
+        out_comm_val = in_comm.copy()
+        out_ost_val = self.xp.zeros_like(in_comm)
```
alfiopuglisi: out_ost_val is overwritten later, no need to allocate it with zero

specula/processing_objects/mmse_piston_unwrapper.py
guidoagapito: Done

specula/processing_objects/mmse_piston_unwrapper.py
guidoagapito: Done

specula/processing_objects/mmse_piston_unwrapper.py
guidoagapito: Done

specula/processing_objects/mmse_piston_unwrapper.py
guidoagapito: Done

alfiopuglisi (approved):

PR #653, opened 2026-05-15
(GA) ERIS parameter files updated

guidoagapito:


config/ERIS/params_eris.yml
```diff
@@ -126,15 +119,16 @@ detector_lgs:
     in_i: 'sh_lgs.out_i'
   outputs:  ['out_pixels']
 
+
 slopec_lgs:
   class:             'ShSlopec'
-  thr_value:         10
+  thr_value:         0
```
alfiopuglisi, line 125: In the real ERIS we do have the threshod, is there a reason to remove it?

config/ERIS/calib_eris_rec_lgs.yml
```diff
@@ -20,14 +21,15 @@ dm_override:
...
+  dt:                0.001
   photon_noise:      False
   readout_noise:     False
+  excess_noise:      False
 
 
 im_calibrator:
   class: 'ImCalibrator'
   nmodes: 700
-  first_mode: 2
```
alfiopuglisi, line 30: if we remove this, the TT will be included in the LGS calibration. Is this the intended behaviour?

guidoagapito: Yes it is intended, the tip-tilt is removed by slicer_lgs.

alfiopuglisi (approved):

config/ERIS/params_eris.yml
guidoagapito: In simulation a threshold of 0 is a reasonable value.

PR #662, opened 2026-05-21
LIFT Sampling

cedricplantet:
Changed PSF model in LIFT to better match the SH model, in particular on sampling aspects.
Added RON as an input in LIFT.
Added the option to have set_total <= 0 in toccd() in order to force the function not to re-normalize (useful for zero-average arrays).


specula/lib/toccd.py
```diff
@@ -82,6 +79,12 @@ def toccd(a, newshape, set_total=None, xp=None):
     eps = xp.finfo(rebinned.dtype).eps
     rebinned_sum = xp.maximum(rebinned.sum(), eps)
 
+    if set_total is None:
+        set_total = a.sum()
+    elif set_total <= 0:
```
alfiopuglisi, line 85: The docstring on set_total should be updated to this new behaviour

specula/processing_objects/lift.py
```diff
@@ -106,6 +109,7 @@ def __init__(self,
         self.ref_tilt = 0.0
         self.padded = None
         self.radians_per_pixel = None
+        self.settings = None
```
alfiopuglisi: This assignment is redundant, self.settings is initalized later

specula/processing_objects/lift.py
```diff
@@ -461,10 +482,29 @@ def crop_or_enlarge_around_peak(self, in_array, desired_width, peak_index=None):
                                     (padding_col, padding_col + extra_col)),
                             'constant', constant_values=0)
         return cropped
+    
+    def _get_tlt_f(self, pupil_size, fft_size):
```
alfiopuglisi, line 487: I think this function could benefit from being cached, like with @lru_cache or similar, since it will be almost always be called with the same arguments

specula/processing_objects/lift.py
```diff
@@ -461,10 +482,29 @@ def crop_or_enlarge_around_peak(self, in_array, desired_width, peak_index=None):
...
+        '''
+        Half-pixel tilt
+        '''
+        iu = complex(0, 1)
```
alfiopuglisi: Is "iu" used in this function?

alfiopuglisi (approved):

PR #666, opened 2026-05-23
Avoid duplicate logging initializations

alfiopuglisi:
This PR fixes a bug that results in multiple log lines when specula.init() is called multiple times. This happens often in tests.


guidoagapito:
I think we should reset logging_initialized global variable before each test.

guidoagapito (approved):

PR #671, opened 2026-05-25
Allow optical gain estimation in open loop

Menessao:
Minor changes to allow the optical gains to be estimated in open loop (i.e. without optical gain compensation on the slopes, simply measuring the optical gain)


specula/processing_objects/optical_gain_estimator.py
```diff
@@ -91,8 +99,12 @@ def _update_optical_gain(self):
         # Avoid division by zero
         if self.xp.abs(demod_cmd) > 1e-12:
             ratio = demod_delta / demod_cmd
-            # Update formula from IDL code
-            updated_gain = current_gain - (1.0 - ratio) * self.gain * current_gain
+
+            if self.open_loop:
+                updated_gain = current_gain - (current_gain - ratio) * self.gain
```
guidoagapito: I would re-write it as:

updated_gain = (1 - self.gain) * current_gain  + self.gain * ratio 

I think that is clearer.

guidoagapito: Another point, in this case we have to check that 0 <= self.gain <= 1 in the init method.

specula/processing_objects/optical_gain_estimator.py
```diff
@@ -12,20 +12,29 @@ class OpticalGainEstimator(BaseProcessingObj):
...
                  #idx_array: list = None, # not supported yet
                  #expression: list = None, # not supported yet
+                 open_loop_estimate: bool = False,
                  target_device_idx: int = None,
                  precision: int = None):
 
         super().__init__(target_device_idx=target_device_idx, precision=precision)
 
+        # Check that integrator gain has sensible values:
+        if gain < 0 or gain > 1:
```
guidoagapito, line 36: 0 and 1 should be included

Menessao: I agree about 0, I was deliberately excluding 1 so that you can simply update with the current value
Let me know if this makes sense, otherwise I'll change to (0,1)

guidoagapito: Sorry, I misunderstood the code, it is fine!

guidoagapito (approved):

alfiopuglisi:
Since this is an interface change (a parameter was added), it is important to update the ChangeLog as well

PR #674, opened 2026-05-27
Removed PhaseScreenCube dependency on pupil stop. 

cedricplantet:
It was useless and provoked a bug where the phase screen was not updating. Indeed the pupistop was set as an input, but it's fixed, so PhaseScreenCube's prepare_trigger was claled just once at the first step of the loop.


guidoagapito:
Ok, I see what you mean. If pupilstop is useful, we can include it as a argument of the init method rather than as input data.

cedricplantet:
> Ok, I see what you mean. If pupilstop is useful, we can include it as a argument of the init method rather than as input data.

It is really useless, we are just acting on the phase in this object. If it turns out useful in the future, we can still add it as you suggested.

guidoagapito (approved):

alfiopuglisi:
Since this is an interface change (an input was removed), it is important to update the ChangeLog as well

PR #684, opened 2026-06-17
Displays with subplots and video recording

alfiopuglisi:
Added options to group multiple displays into a single window, using matplotlib subplot syntax.
Added video recording of displays into mp4 files. Video recording is triggered by sending one or more display outputs into a video recorder object as an input 
Please note the additional dependency (imageio-ffmpeg)


specula/display/display_recorder.py
```diff
@@ -0,0 +1,31 @@
+from specula.display.base_display import BaseDisplay
```
guidoagapito: I think that this recorder should not inherit base_display, it is not a display.

guidoagapito:
In addition to the inline comment, I think that this new features should be detailed in our documentation.

guidoagapito (approved):

PR #685, opened 2026-06-17
(GA) enabled start and end time in field_analyser with tests

### LINKED ISSUE #683: Saving reduced phase cubes
The FieldAnalyser doesn't seem to take into account the start_time and end_time keywords when using compute_phase_cube(). This would be convenient to save a subset of phase screens.

guidoagapito:
This pull request should fix #683.

### DIFF
--- a/specula/simul.py
+++ b/specula/simul.py
@@ -788,7 +788,7 @@ class Simul():
                 else:
                     raise ValueError(f"Invalid number of parts detected in override: {parts}. Did you add/forget a '.'?")
     
-    def run(self):
+    def run(self, start_time=0, end_time=None):
         params = {}
         # Read YAML file(s)
         self.logger.info('Reading parameters from ' + self.param_files[0])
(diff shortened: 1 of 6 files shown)


specula/simul.py
```diff
@@ -851,8 +851,11 @@ def run(self):
             self.loop.add(disp, idx+1)
 
         # Run simulation loop
-        self.loop.run(run_time=self.mainParams['total_time'],
+        total_time = self.mainParams['total_time']
+        run_time = (end_time if end_time is not None else total_time) - start_time
```
alfiopuglisi, line 855: I think this expression is wrong: run_time should be equal to total_time, when end_time is None

alfiopuglisi (changes requested):

specula/simul.py
guidoagapito: The while condition in the LoopControl run method is:

        `while self.run_time < 0 or self.t < self.t0 + self.run_time:`

In LoopControl `run_time` is the time duration from `t0`, not the end time.

alfiopuglisi: Right, then go ahead with the merge

alfiopuglisi (approved):

PR #686, opened 2026-06-18
Simple DM command clipping logic

Menessao:
Added the stroke parameter to the DM class, allowing to set stroke limitations on modal or zonal commands.
Stroke can be a scalar or a list of values.


specula/processing_objects/dm.py
```diff
@@ -156,7 +179,8 @@ def input_names(cls):
 
     @classmethod
     def output_names(cls):
-        return {'out_layer': OutputDesc(Layer, 'Output wavefront layer produced by the DM')}
+        return {'out_layer': OutputDesc(Layer, 'Output wavefront layer produced by the DM'),
+                'out_clip_command': OutputDesc(Layer, 'DM applied command, after (optional) clipping')}
```
alfiopuglisi: Not a Layer, it's a BaseValue

specula/processing_objects/dm.py
```diff
@@ -1,6 +1,6 @@
 
 from specula.base_value import BaseValue
-from specula.connections import InputValue
+from specula.connections import InputValue, Output
```
alfiopuglisi: Is this Output needed?

specula/processing_objects/dm.py
alfiopuglisi: should also be "out_clipped_command" to be consistent with the output definition

specula/processing_objects/dm.py
```diff
@@ -177,6 +204,8 @@ def trigger_code(self):
             self.layer.phaseInNm[self._ifunc.idx_inf_func] = \
                 self.if_commands[self.if_commands_selector] @ self._ifunc.influence_function[self._valid_modes, :]
         self.layer.generation_time = self.current_time
+        self.clip_command.value = cmd
```
alfiopuglisi: It's best to use a syntax like value[:] = cmd, so that the value array will be updated in place, on the GPU it is more efficient

specula/processing_objects/dm.py
```diff
@@ -69,7 +70,11 @@ def __init__(self,
         pupilstop : Pupilstop
             Pupilstop object defining the DM aperture.
         sign : int [1], optional
-            Sign for the DM surface deformation, by default -1 (to account for reflection).
+            Sign for the DM surface deformation, by default -1.
+        stroke : float or list, optional
```
guidoagapito: you should add in square brakets the unit, [nm]

guidoagapito:
Please update ChangeLog.md

guidoagapito (approved):

PR #687, opened 2026-06-18
(GA) PhaseExtractor class with test

guidoagapito:


alfiopuglisi:
The new object is fine, but why is the extracted phase flattened? Can't we keep it 2d as it originally?

guidoagapito:
> The new object is fine, but why is the extracted phase flattened? Can't we keep it 2d as it originally?

You are right, we should keep the original shape, I will update the pull request.

alfiopuglisi (approved):

PR #690, opened 2026-06-19
CLOSE gain optimizer

Menessao:
Correlation based algorithm for modal gain optimization of integrator controllers, with relevant tests.

The reference paper is "CLOSE: a self-regulating, best-performance tracker for modal integrator based AO loops" by Deo et al. 2019, which describes a correlation based method to adapt integrator gains in order to obtain a given amplification at the integrator resonance. It clearly has nothing to do with the optical gain estimation method using the high frequency part of the PSD described in "Shack-Hartmann wavefront sensor sensitivity loss factor estimation in partial correction regime" by Agapito et al. 2017


alfiopuglisi (approved):

guidoagapito (approved):

Menessao:
You know I have used Gemini and Co-pilot when all tests pass first try and I get a double approval with no changes requested

PR #691, opened 2026-06-22
Fresnel propagation update

bernadett92:
To be more accurate and stable:
- added Fraunhofer propagation for far-field
- checked limits for FFTs and ASM method (propagation distance is reduced automatically)
- avoid zero padding/unpadding in propagation between layers


specula/processing_objects/atmo_propagation.py
```diff
@@ -32,6 +32,9 @@ def __init__(self,
                  mergeLayersContrib: bool=True,
                  upwards: bool=False,
                  padding_factor: int=1,
+                 seeing: float = -1.0,
```
alfiopuglisi: We already have the seeing as an input for AtmoEvolution, usually from some kind of generator, it should be the same here: an optional input of type BaseValue, instead of an init parameter

specula/processing_objects/atmo_propagation.py
```diff
@@ -220,7 +257,51 @@ def _build_source_heights(self):
...
+                "For the current configuration the propagation distance is too large."
+                "Thus it is reduced from " + str(z_in) + "m to " + str(z) +
+                "m. Consider increasing zero padding.")
+
+        self.z_total += z
+
+        return propagator
+
     def doFresnel_setup(self):
+        if self.local_inputs['seeing'] is not None:
```
guidoagapito: I suggest adding `self.z_total = 0` at the beginning pf this method

specula/processing_objects/atmo_propagation.py
```diff
@@ -220,7 +257,51 @@ def _build_source_heights(self):
...
+            z_max_ASM = self.ef_size_padded * self.pixel_pitch ** 2 / (self.wavelengthInNm * 1e-9)
+            z = min(z, z_max_ASM)
+            propagator = self.asm_propagator(z, self.pixel_pitch, self.pixel_pitch)
+        else:
+            propagator, _, _ = self.fraunhofer_propagator(z, self.pixel_pitch)
+
+        if z != z_in:
+            self.logger.warning(
+                "For the current configuration the propagation distance is too large."
+                "Thus it is reduced from " + str(z_in) + "m to " + str(z) +
```
guidoagapito, line 288: add a space before `Thus`

specula/processing_objects/atmo_propagation.py
```diff
@@ -220,7 +257,51 @@ def _build_source_heights(self):
...
+                "m. Consider increasing zero padding.")
+
+        self.z_total += z
+
+        return propagator
+
     def doFresnel_setup(self):
+        if self.local_inputs['seeing'] is not None:
+            seeing = self.local_inputs['seeing'].value[0]
```
guidoagapito: Maybe `float(self.local_inputs['seeing'].value)` instead of `[0]`?

specula/processing_objects/atmo_propagation.py
```diff
@@ -335,17 +412,23 @@ def trigger_code(self):
                     self.ef_temp.phaseInNm[:] = self.prop_sign * interpolator.interpolate(layer.phaseInNm)
 
                 if self.doFresnel:
-                    self.ef_fresnel *= self.ef_temp.ef_at_lambda(self.wavelengthInNm)
+                    self.ef_fresnel[s:s + self.pixel_pupil, s:s + self.pixel_pupil] *= self.ef_temp.ef_at_lambda(
+                        self.wavelengthInNm)
                     if self.propagators[li] is not None:
-                        self.angular_spectrum_propagation(self.ef_fresnel, self.propagators[li])
+                        if type(self.propagators[li]) == list:
```
guidoagapito: Please use `isinstance`, or even better use a flag to store the ASM / fraunhofer case.

specula/processing_objects/atmo_propagation.py
```diff
@@ -282,25 +359,24 @@ def prepare_trigger(self, t):
...
-        self.ef_padded[s:s + self.pixel_pupil, s:s + self.pixel_pupil] = ef_in
+    def fraunhofer_far_field_propagation(self, ef_in, propagator):
+        self.ft_ef1[:] = self.xp.fft.fft2(self.xp.fft.fftshift(ef_in, axes=(-2, -1)), axes=(-2, -1),) * self.pixel_pitch ** 2
+        self.ef_fresnel[:] = propagator * self.xp.fft.fftshift(self.ft_ef1, axes=(-2, -1))
 
+    def angular_spectrum_propagation(self, ef_in, propagator):
         if propagator[0] is not None:
-            self.ef_padded *= propagator[0]
-        self.ft_ef1[:] = self.xp.fft.fft2(self.xp.fft.fftshift(self.ef_padded, axes=(-2, -1)), axes=(-2, -1),
+            ef_in *= propagator[0]
```
guidoagapito, line 360: This is done in place, is that how it is supposed to be? Previously this multiplication is done on `ef_padded`.

specula/processing_objects/atmo_propagation.py
```diff
@@ -156,9 +169,33 @@ def __init__(self,
...
+
+       Parameters
+       ----------
+       distanceInM : float [m]
+           Propagation distance in meter.
+       d_in : float [m]
+           Grid spacing in the source plane
+       """
+        k = 2 * np.pi / (self.wavelengthInNm * 1e-9)
+        self.d_out = (self.wavelengthInNm * 1e-9 * distanceInM) / (self.ef_size_padded * d_in)
```
guidoagapito: Should `d_out` be a member of self? Or `d_out` is enough? I think that it is not used anywhere else.

guidoagapito: Maybe it must be used in the trigger:

`trigger_code()` always extracts `ef_fresnel[s:s+pixel_pupil, s:s+pixel_pupil]` and assigns it to `output_ef`, which still has a pitch of `pixel_pitch`, ignoring the fact that after the Fraunhofer line the grid has a pitch of `d_out`. If `d_out != pixel_pitch`, the spatial correspondence between pixels and metres in the output field is incorrect.

specula/processing_objects/atmo_propagation.py
guidoagapito: or after `large.`

specula/processing_objects/atmo_propagation.py
```diff
@@ -182,19 +177,24 @@ def fraunhofer_propagator(self, distanceInM, d_in):
...
+        H_FR = (self.pixel_pitch ** 2 / (1j * self.wavelengthInNm * 1e-9 * distanceInM)) * self.xp.exp(
+            -1j * (k / (2 * distanceInM)) * (x_in_out ** 2 + y_in_out ** 2))
 
-        return H_FR, x_out, y_out
+        # Chirp Z-transform for performing DFT
+        row_idx = self.xp.arange(self.ef_size_padded) - (self.ef_size_padded // 2) + 0.5
+        col_idx = self.xp.arange(self.ef_size_padded) - (self.ef_size_padded // 2) + 0.5
+        H_FR_kernel = self.xp.exp(
+            -1j * k / distanceInM * self.xp.outer(row_idx * self.pixel_pitch, col_idx * self.pixel_pitch))
+        H_FR_in = self.xp.exp(1j * k / (2 * distanceInM) * (x_in_out ** 2 + x_in_out ** 2))
```
guidoagapito: There is two times `x_in_out ** 2`

specula/processing_objects/atmo_propagation.py
```diff
@@ -32,6 +33,7 @@ def __init__(self,
                  mergeLayersContrib: bool=True,
                  upwards: bool=False,
                  padding_factor: int=1,
+                 beam_center=[0,0],
```
guidoagapito: `beam_center=[0, 0]` uses a mutable list as a default argument, which is a Python anti-pattern: all callers that rely on the default share the same object, so any in-place mutation of `self.beam_center` (e.g. via `self.xp.array(beam_center)`) would not affect them in this case, but a future `beam_center.append(...)` elsewhere would silently corrupt every instance. The idiomatic fix is:

```
def __init__(self, ..., beam_center=None, ...):
    ...
    self.beam_center = self.xp.array(beam_center if beam_center is not None else [0, 0])
```

specula/processing_objects/atmo_propagation.py
```diff
@@ -74,6 +76,10 @@ def __init__(self,
             (downwards).
         padding_factor : int [1], optional
             Factor for zero padding in Fresnel propagation to avoid numerical issues with FFTs.
+        beam_waist : float [m], optional
```
guidoagapito: `beam_waist` was removed in the last commit

guidoagapito (dismissed):

specula/processing_objects/atmo_propagation.py
```diff
@@ -583,7 +661,8 @@ def setup(self):
         self.setup_interpolators()
         if self.doFresnel:
             self.doFresnel_setup()
-        self.build_stream()
+        else:
```
guidoagapito, line 664: Is that a deliberate choice?

specula/processing_objects/atmo_propagation.py
```diff
@@ -74,6 +76,8 @@ def __init__(self,
             (downwards).
         padding_factor : int [1], optional
             Factor for zero padding in Fresnel propagation to avoid numerical issues with FFTs.
+        beam_center : float [m], optional
```
guidoagapito: Is the unit of measurement metres or pixels? From the code, it looks as though it is in pixels.

guidoagapito:
Please update ChangeLog.md

specula/processing_objects/atmo_propagation.py
bernadett92: I did this on purpose. Otherwise I got an error with 'calling cuBLAS API during stream capture is currently unsupported' at the matrix multiplication in the Fraunhofer propagation at 
self.ft_ef1[:] = propagator[2] @ (ef_in * propagator[1]) @ propagator[2].T

But maybe there is a better way to do it.

guidoagapito (approved):

PR #693, opened 2026-07-02
(GA) avoid using a never generated input in BaseOperation

guidoagapito:


specula/processing_objects/base_operation.py
```diff
@@ -219,8 +219,13 @@ def setup(self):
     def trigger_code(self):
 
         value1 = self.local_inputs['in_value1'].value
+        if self.local_inputs['in_value1'].generation_time < 0:
+            value1 = value1*0.
```
alfiopuglisi: This will fail if the unititialized value1 contains NaNs. I think it's better to use zeros_like(value1) or similar

specula/processing_objects/base_operation.py
```diff
@@ -219,8 +219,13 @@ def setup(self):
...
+        value2 = None
```
alfiopuglisi: There should be no need for this initalization, value2's validity is checked in setup()

specula/processing_objects/base_operation.py
```diff
@@ -219,8 +219,13 @@ def setup(self):
...
         if self.local_inputs['in_value2'] is not None:
-            value2 = self.local_inputs['in_value2'].value 
+            value2 = self.local_inputs['in_value2'].value
+            if self.local_inputs['in_value2'].generation_time < 0:
+                value2 = value2*0.
```
alfiopuglisi: Same comment as value1*0

specula/processing_objects/base_operation.py
guidoagapito: Done

specula/processing_objects/base_operation.py
guidoagapito: Done

specula/processing_objects/base_operation.py
guidoagapito: Done

alfiopuglisi (approved):

PR #695, opened 2026-07-06
Unclear usage of override

jodlbauer:
The overrides option is a bit confusing.
https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/b024ca793ba45267a8f401185c39f617d94d01bc/specula/simul.py#L55-L58
looks like overrides needs to be a list, but this leads to failure in the yaml parser. Using a yaml string directly (e.g. `overrides="main.total_time: -1"`) works.

Is this the intended usage? At first glance it looked like a for loop over the overrides is missing here:
https://github.com/ArcetriAdaptiveOptics/SPECULA/blob/b024ca793ba45267a8f401185c39f617d94d01bc/specula/simul.py#L777-L780

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -29,6 +29,7 @@
 - Fixed output\_names in PhaseScreenCube
 - Fixed PhaseScreenCube crash on GPU due to np.searchsorted called on a cupy array
 - Fixed start\_time bug in WindowedIntegration
+- Fixed unclear/inconsistent overrides default value and type hint in Simul
 - Optimization of the compute\_ifs\_covmat function
 - ...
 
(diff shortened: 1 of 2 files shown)


guidoagapito:
Looks correct — overrides is used as a single YAML-mapping string everywhere else (apply_overrides's yaml.full_load(self.overrides), main_simul's overrides: str, and the --overrides CLI arg), so [] was never the right default type, just one that happened not to crash. "" fixes that.

On the missing-for-loop question: there isn't one missing — yaml.full_load(self.overrides).items() already iterates over every key in the mapping, so multiple overrides go in one string (e.g. "main.total_time: -1, foo.bar: 2"), not a list of strings.

guidoagapito:
Testa are failing on Python 3.8 and 3.9, I think you need to add:

`from __future__ import annotations`

in simul.py

guidoagapito (approved):

Issue #696, opened 2026-07-06
FieldAnalyser doesn't seem to take into account EF combinators


cedricplantet:
I have seen an issue when extracting phase cubes from simulations where I have user defined pase screens added in input (via PhaseScreenCube and EF combinator). What I see in the resuting phase is approximately the input phase screen with opposite sign. Indeed, FieldAnalyser seems to only consider what is in the propagation object, whereas my phase screens were added via the EF combinator. In the end, we only get the sum of the atmosphere and the correction, which tends towards minus the input phase screen. I checked that by running a MORFEO simulation with an additional constant offset of tip of 1 um rms on all sensors, and indeed I get a cube tending to a tip of - 1 um rms.

guidoagapito:
#702 and #703 were merged and should solve this issue.

PR #702, opened 2026-07-22
(GA) FieldAnalyser looks for dropped EF with tests

guidoagapito:


guidoagapito:
FieldAnalyser will generate an error in the case of issue #696

specula/field_analyser.py
```diff
@@ -44,6 +45,7 @@ def __init__(self,
         self.start_time = start_time
         self.end_time = end_time
         self.display = display
+        self.on_missing_downstream_consumers = on_missing_downstream_consumers
```
alfiopuglisi, line 53: Here we should check that on_missing_downstream_consumers has one of the allowed values ('error', 'warn' etc) and not something else

guidoagapito: Done

alfiopuglisi (approved):

PR #705, opened 2026-07-29
Generate yaml json-scheme from __init__ methods

jodlbauer:
Hi, I always found it a bit annoying having to look in the source code to figure out which parameters certain classes need, so I played around a bit with the yaml json-scheme. Maybe this can be interesting for others as well.

- Can be used to get autocomplete in parameter files using e.g. VSCode (Jetbrains yaml parser acts a bit weird) by putting
`# yaml-language-server: $schema=./specula.schema.json`
 as first line into parameter files with the correct path. Alternatively most IDEs can be told which scheme to use.

- ~~There are two implementations, both do essentially the same:~~
  - ~~`make_scheme.py` using inspect~~
  - ~~adjusted the `parse_classes.py`~~
~~(I started with `inspect`, then tried to get it into the existing `parse_classes.py`)~~

- Uses typing information, fallback to ~~string~~ number.
- ~~Currently only running on the classes defined in `parse_classes.py`, others should work as well.~~
- ~~specifiers like _ref not yet supported~~


jodlbauer:
Rebased to main

guidoagapito:
Hi @jodlbauer, thanks for this: a JSON schema for YAML autocomplete/validation is a useful idea, and it's not a duplicate of what we already generate for docs. In fact, `docs/scripts/generate_objects_summary.py` builds human-readable RST documentation from the same classes, not a schema for editors.

Before going any further, however, I would like to clarify a few points:

- `make_scheme.py` (inspect-based) and the changes in `parse_classes.py` (AST-based) do almost the same job — mapping Python types to JSON schema — with near-duplicate code. Could you settle on a single approach and drop the other? We'd lean towards the AST-based one already in `parse_classes.py`, since it parses the source without importing/executing the module, which is safer and consistent with how the rest of the class-parsing already works.

- `exposed_classes` only lists ~15 classes by hand, so most processing/data objects won't get a schema entry. `generate_objects_summary.py` already discovers all classes automatically by scanning the package directories — could the schema generator reuse that same discovery mechanism instead of a manual list? That keeps it complete as new classes get added, with no extra maintenance.

- `specula.schema.json` is committed by hand, so it can silently drift out of sync with the code whenever a class changes and nobody remembers to regenerate it. Our existing docs generation instead runs automatically as part of the docs build (see `docs/conf.py`), never committed. Could the schema follow the same pattern — generated at build/test time, or at least checked in CI that it's up to date?

- You note `_ref` isn't supported yet. Since SPECULA yaml configs often reference other sections, the schema may flag valid configs as invalid until this is handled — worth resolving, or at least clearly documented as a known limitation, before merge.

In general, I'd rather not end up with two parallel ways of extracting the same class info — this code could be incorporated into the existing `parse_classes.py` file (and the detection logic contained in the `generate_objects_summary.py` file could be reused), so as to have a single point of reference.

jodlbauer:
- make_scheme: that was the first draft with inspect, since I was more familiar with that module, and ast was rather new for me. I only discovered the parse_classes afterwards. But yes, one of those implementations will be removed.
- I didn't know about the generate_objects_summary, will look into that.
- Committing the schema itself: mostly to make it easier for you to quickly try it, but also because the question remains where to put it so that the users can point their IDE to that file. Do you suggest that the CI should create that file, or that it is created on pip install?
- ref is a bit of a challenge, since it is not a parameter in init. I could propably add a _ref version for each parameter, which at least avoids marking valid files as invalid (there will still be misc. warnings, since the typing information is sometimes not correct)

guidoagapito:
On the `_ref` support: it's actually a bit more general than just `_ref`. There's a single generic mechanism in `simul.py` that strips one of several suffixes off any parameter name before matching it against the constructor:

```
for ending in ['_ref', '_data', '_object']:
    candidate = remove_suffix(parname, ending)
    if candidate in args:
        parname = candidate
        break
```

So for basically any parameter, users can write `X` (inline value), `X_ref` (name of another object defined in the same yaml file), `X_data` (a tag, data read from a FITS file), or `X_object` (a tag, a full object restored from a FITS file). For parameters typed as List/Dict there are also `X_list_ref`, `X_dict_ref`, `X_list_object`, `X_dict_object`. This isn't specific to `ifunc/simul_params` — it applies to any parameter of any class. So whatever fix goes in for `_ref` will need to cover `_data` and `_object` too (and the list/dict variants), otherwise plenty of valid files will still get flagged as invalid.

One suggestion, take it or leave it: rather than generating `X`, `X_ref`, `X_data`, `X_object` (and the list/dict versions) as separate explicit properties for every single parameter — which means knowing in advance which suffixes apply to which parameter based on its type — you could add a `patternProperties` block once per class (or even once globally):

```
"patternProperties": {
  "^\\w+_(ref|data|object)$":      { "type": "string" },
  "^\\w+_(list_ref|list_object)$": { "type": "array", "items": {"type": "string"} },
  "^\\w+_dict_(ref|object)$":      { "type": "object", "additionalProperties": {"type": "string"} }
}
```

With `additionalProperties: false`, any key matching one of these patterns is still accepted, so this handles all suffix forms in one shot without enumerating them per parameter. The trade-off is you lose some autocomplete precision (the IDE suggests "any key ending in `_ref`" rather than `simul_params_ref` specifically) and it gets harder to enforce required on a parameter that could legitimately be satisfied by any of several suffixed forms — probably simplest to only keep required for plain scalar parameters and drop it for the object/reference-typed ones. Just an idea to keep the generator simpler; the `_data`/`_object` gap above is the one that actually needs fixing.

guidoagapito:
I forgot the "committing the schema" point.

I would lean towards not committing it, and instead generating it in CI with a check that fails if regenerating produces a diff (same idea as some projects do for generated protobuf/graphql schemas) — that keeps it always in sync without relying on someone remembering to regenerate by hand.
I would not tie it to `pip install`: that needs a custom build/post-install hook (fragile with editable installs, `--no-build-isolation`, etc.), and it does not really solve the discoverability problem anyway — the `yaml-language-server` directive needs a path relative to the user's own parameter files, not to site-packages, so users would still need to fetch/copy it locally regardless of where `pip` puts it. Keeping it as a script (like today) that users/CI can run on demand, with CI enforcing freshness, seems like the simplest solution that fits how `generate_objects_summary.py` already works for docs.

jodlbauer:
I think I have got it now in a quite good state. It supports loading from a `tag`, and the postfixes `_ref`, `_object`, `_data` (is _dict_ref a separate postfix, or is the _dict part of the name?).
I couldn't get it working properly with the `patternProperties`, so for now I added all of the postfixes as properties, and then defined that only one of those can be used (or none).
Downside is that the scheme is now at 20k loc, but seems to be still ok in VSCode.

There is a small issue due to the way the AST parser works:
if inputs/outputs are defined in a parent class, the AST does not return them for the child class, so the scheme does not know about those. This would be nice to fix, but probably adds quite some complexity to the parser if it has to trace back the inheritance tree.

There are also some (partially) incorrect typing info: e.g. `constant` from `WaveGenerator` is typed as `list[float]`, but also accepts `float` (but I think this is outside the scope of this PR). I just fixed one obvious mistake in this PR now to prevent some errors.

guidoagapito:
Thanks, this is in good shape.

On `_dict_ref` / `_list_ref`: they are not independent postfixes.
The stripping only removes `_ref`/`_data`/`_object`;
`_dict`/`_list` are part of the parameter own name by SPECULA convention (e.g. `source_dict`, `atmo_layer_list`).
So you do not need separate postfix categories for those — just `_ref`/`_data`/`_object` on the parameter's real name, with the accepted value type (string vs array vs dict of strings) driven by that parameter own type hint, which you already extract.

On inherited inputs/outputs not showing up: that is an expected limitation of static AST parsing, not a bug — worth a one-line note in the docs so users know why some inherited I/O might be missing, but nothing to fix here.

On imprecise type hints (e.g. `WaveGenerator.constant`): that one is a real issue, just in the SPECULA source itself rather than in your tool — the type hint is narrower than what the code actually accepts.
Worth opening a separate issue to track and clean up over time;
good call fixing the one obvious case here without trying to solve them all in this PR.

No objection to dropping `patternProperties` if it did not work out cleanly — explicit properties are more verbose but more precise, and if VSCode still handles it fine at this size, that is a reasonable tradeoff.

FabioRossiArcetri:
I am finally looking into this today

FabioRossiArcetri (approved):

guidoagapito:
I do not consider the work on this pull request concluded and it is still a draft pull request.

guidoagapito:
@jodlbauer I’m sorry, this pull request was merged by mistake and we have reverted it. You will now need to create a new pull request to continue with development.

PR #706, opened 2026-07-29
Avoid phase wrapping in Fresnel propagation

Menessao:
Simple trick to avoid phase wrapping during Fresnel propagation
Plus some test suggested by Gemini, with some adjustements


Menessao:
I was not able to add @bernadett92 among the reviewers ...

guidoagapito:
Hi @Menessao, thanks for looking into this: phase wrapping in the Fresnel path is a real thing to think about, and I appreciate the tests you added around it.

That said, I don't think this is the right fix. With real long-path atmospheric propagation, doFresnel is exactly the regime where scintillation shows up, and phase wrapping there isn't just a numerical artifact, it reflects genuine amplitude nulls/vortices in the field, which no amount of incremental tracking can properly resolve. The current approach can end up looking "unwrapped and correct" while quietly being wrong in exactly the cases we care about most.

We actually already have a proper solution for this, just applied in a different (and I think better) place: `ModalAnalysis.unwrap_ls` does a real 2D spatial phase unwrap (least-squares/Poisson method), validated against `skimage.restoration.unwrap_phase` in `test_modal_analysis.py`, and there's already a test (`test_modal_analysis_unwrapping_physical_propagation`) that exercises exactly this scenario: `doFresnel=True` producing a wrapped phase, unwrapped downstream by `ModalAnalysis` when needed. So the wrapped output from `AtmoPropagation` isn't an oversight, it's the existing design, and unwrapping is meant to happen where it's actually needed, using a tested method, rather than at the source for everyone.

Changing `AtmoPropagation` to pre-unwrap via incremental tracking risks interfering with that existing mechanism and adds a third, ad hoc way of unwrapping phase in the codebase alongside the two we already have.

I'd suggest closing this one. If we think the wrapped-phase contract of `AtmoPropagation` needs to be clearer/documented, that's worth its own issue, but not this fix. Happy to discuss further if useful.

Menessao:
@guidoagapito I tried the phase unwrapping with the ModalAnalysis and it was not working well for me, but it is entirely possible that I was simply using it wrong.

The reason I am requesting this change is because I tried repeating a RISTRETTO-like simulation with doFresnel = True. On an 8m class telescope, I expect scintillation to affect the contrast a bit, but barely change the Strehl. However, before this modification, I observed a drop in Strehl from 80% to 30% in my simulation, which is rectified with this change. Contrast is still a bit affected.

If you think this is not an issue or that I simply must have made a mistake (which is not outside the realm of possibilities), I am ok with closing the pull request.

Let me know

guidoagapito:
> @guidoagapito I tried the phase unwrapping with the ModalAnalysis and it was not working well for me, but it is entirely possible that I was simply using it wrong.
> 
> The reason I am requesting this change is because I tried repeating a RISTRETTO-like simulation with doFresnel = True. On an 8m class telescope, I expect scintillation to affect the contrast a bit, but barely change the Strehl. However, before this modification, I observed a drop in Strehl from 80% to 30% in my simulation, which is rectified with this change. Contrast is still a bit affected.
> 
> If you think this is not an issue or that I simply must have made a mistake (which is not outside the realm of possibilities), I am ok with closing the pull request.
> 
> Let me know

Quick question that could explain the SR drop: in your RISTRETTO-like setup, is the `wavelengthInNm` you pass to `AtmoPropagation` the same as the one you pass to PSF?

The Fresnel path derives `phaseInNm` from `angle(ef) * wavelengthInNm / (2 pi)`, which wraps it modulo the propagation wavelength specifically. If PSF then re-expands that phase at a different wavelength, you'd get large spurious phase errors purely from that mismatch, which could easily produce an 80% --> 30% SR drop on its own, with nothing to do with real scintillation.

`ElectricField` already has a wavelengthInNm tag + tolerance check (`_check_wavelength_compatibility`) that raises a clear error if a field is evaluated at the wrong wavelength. It looks like `AtmoPropagation` Fresnel output isn't using it. Tagging that output field with its propagation wavelength would catch this kind of mismatch loudly instead of silently corrupting the Strehl.

bernadett92:
I agree with @guidoagapito and also think that this is not the right fix as we care in many settings about the changed phase with Fresnel propagation, which implies wrapping and phase singularities.

Another thing that could explain the Strehl ratio drop is that phase singularities (branch points) occur, where modal reconstructors do not perform very well. There is e.g. a paper by Fried explaining this:
https://opg.optica.org/viewmedia.cfm?r=1&rwjcode=josaa&uri=josaa-15-10-2759

Menessao:
Thanks for commenting and providing the source, @bernadett92

The phase-screens and seeing considered in my simulation would give me a Rytov number of about 0.02 (indeed, I am considering an astronomical HCI application).
While I would completely agree with you if my simulations were in the strong scintillation regime, should I still expect phase wrapping in this weak scintillation case?

guidoagapito:
@Menessao, a valid point on the Rytov number: at about 0.02 you are in the weak-scintillation regime, so I would not expect branch points to be the main driver in your case either.

I notice you have also pushed some updates to the PR since then, but I do not think they answer the question I asked above: is the `wavelengthInNm` passed to `AtmoPropagation` the same as the one passed to PSF in your setup?
That is a much simpler explanation for an 80% --> 30% SR drop, unrelated to scintillation.

Could you answer that, and share the parameter file you used?
That would let us reproduce it directly and understand the cause quickly, rather than iterating further.

Menessao:
It was indeed not the same value: I was not giving any wavelength in input to AtmoPropagation, so it was using the default 500 nm, which differs from the 750 nm I was evaluating the PSF at.
Having updated it now to 750 nm, I am unable to close the loop (with the same filter used to get 80% SR) unless unwrapping the phase.
The issue seems to be fixed when operating the pyramid at the same wavelength as the source and propagation wavelength, which I did not expect to be necessary, as this does not matter for geometric propagation.

You should find the relevant configuration file attached.

[ristretto_bright.yml](https://github.com/user-attachments/files/30579661/ristretto_bright.yml)

guidoagapito:
Ok, this confirms the mismatch was real.

Looking at `ristretto_bright.yml` file: `prop` does physical (Fresnel) propagation at 750nm, and its output field feeds both `pyr` (which operates at 1200nm) and `psf1`/`coro_psf1` (at 750nm).
Physical/Fresnel propagation is chromatic: diffraction (not just phase, the amplitude/scintillation structure too) depends on wavelength.
So a field propagated at 750nm is not physically valid input for a 1200nm sensor, you'd need a second `AtmoPropagation(doFresnel=True, wavelengthInNm=1200)` feeding the pyramid path, and keep the 750nm one only for the coronagraph/PSF path.

Right now nothing in SPECULA stops you from wiring it up this way: there is no check that a Fresnel-propagated electrical field is being used by a processing object at the wavelength it was actually computed for.
We should add this check (ElectricField already has a wavelength-tagging + tolerance check mechanism that could be reused here) so this kind of mismatch raises a clear error instead of silently giving wrong results.

Could you try splitting the propagation into two (1200nm for the pyramid, 750nm for the PSF) and see if you can close the loop at 80% SR without any unwrapping?

Menessao:
I checked and indeed having two different propagations allows to achieve the expected performance. The residuals are still wrapped, of course, which might not be expected for ExAO cases (and unwrapping in ModalAnalysis does not help, unless you do not care about fitting error). I attach sample residual phases for both cases, current and after this PR's "fix".

image image

I can add the check on the wavelength, but should I always raise an error if the electric field wavelength is not the same as the WFS wavelength? When using geometric propagation I don't think this matters and it probably should not raise an error. 
Is there a way to know if an electric field was Fresnel-propagated? Should I add a tag or a check on the amplitude?

guidoagapito:
Glad the two-propagation split gets you back to the expected performance: that confirms the diagnosis.

On your question: yes, I would tag the wavelength if and only if the field went through Fresnel propagation, and leave geometric fields untagged (as today).
Checking the amplitude to guess whether Fresnel was used is not reliable either way: under weak scintillation the amplitude stays close to uniform even though the field is genuinely wavelength-locked, and non-uniform amplitude can also come from other things (masks, coronagraphs) that have nothing to do with a wavelength constraint.
Better to have `AtmoPropagation` say so explicitly, since it is the one that actually knows.

The mechanism for this already exists in `ElectricField`: `wavelength_in_nm` defaults to `None` (no constraint, which is correct for geometric propagation), and `_check_wavelength_compatibility` only raises when it is set.
So the change is small: in the `doFresnel=True` path, set `output_ef.wavelength_in_nm = self.wavelengthInNm` once at setup, and leave the geometric path untouched.

On the still-wrapped residuals for ExAO / fitting error: that is a real and separate question worth its own discussion, but it does not change the diagnosis here.
One part of it is the unwrapped DM phase being summed with the wrapped atmo phase in the combinator, another part is the high-spatial-frequency residual content that the unwrap does not preserve well.

I'd suggest closing this PR: the original 80% --> 30% drop is explained by the wavelength mismatch, and the unwrap approach is not the right fix for it.
Could you open a new, separate PR for the wavelength-tagging safeguard we discussed?
It is a small, self-contained change and deserves its own review rather than being folded into this one.

PR #708, opened 2026-08-04
Added wavelength compatibility checks for Fresnel propagation

Menessao:


specula/processing_objects/atmo_propagation.py
```diff
@@ -166,6 +167,12 @@ def __init__(self,
 
         self.airmass = 1. / np.cos(np.radians(self.simul_params.zenithAngleInDeg), dtype=self.dtype)
 
+        if self.doFresnel:
```
guidoagapito: `if self.mergeLayersContrib == False` then `'out_'+source_name+'_ef'` does not exist here. So, please, move this in the setup method.

specula/data_objects/electric_field.py
```diff
@@ -212,6 +212,7 @@ def ef_at_lambda(self, wavelengthInNm, slicey=None, slicex=None, out=None):
             The electric field at the given wavelength. If out is provided, the result is
             both stored in out and and the same reference is returned.
         """
+        self._check_wavelength_compatibility(wavelengthInNm)
```
guidoagapito: this is not required, at line 219 we have 

`phi = self.phi_at_lambda(wavelengthInNm, slicey=slicey, slicex=slicex)`

which does this check

guidoagapito (approved):

PR #710, opened 2026-08-06
Pyramid tilt coefficients + minor size bugfix

Menessao:
Implemented the pyramid tilt coefficients to allow different tilts on the pyramid faces, enabling simulation of 'imperfect' pyramids (e.g. EKARUS's pyramid)

Also added a minor fix on the pyramid mask size, discovered by mistake: when using a pupil mask with a large and odd number of pixels, it could happen that the pyramid masks(pyr_tlt and fp_mask) had incompatible shapes, raising an error


Menessao:
I am opening as draft because:
- I am unsure if I implemented pyr_tlt_coeffs as they were intended
- the tests I developed are quite time consuming, and might need to be sped up a bit

guidoagapito:
Nice work, this is close. The `//2` -> `ceil` fix is well targeted and the regression test for odd pupil sizes is a good addition.

On the tilt implementation: good that with all coefficients set to 1 it reduces exactly to the existing "untilted" case in the `else` branch — worth adding an explicit test for that (`pyr_tlt_coeff` all-ones vs `None` should give identical output), cheap safety net for the future.

Two things worth fixing before merging:

- `pyr_tlt_coeff` is typed (and documented) as `float`, but it's actually a 2x4 array. Could you fix the type hint/docstring, and add a basic shape check with a clear error message? Right now a wrong shape would fail with a cryptic index/broadcasting error.

- In `test_pyramid_tilt_coeffs`, the tilt is applied to face index 0, but the assertion checks the deviation of the last detected pupil (`[-1]`), and the naming looks swapped (`k[0,i]` multiplies `x`, but the test assigns `tlt_y` to it). This might well be correct if the focal-plane quadrant order doesn't map 1:1 to the detector pupil order, but could you confirm you checked this mapping directly rather than assumed it?

A minor point: `self.xp.ceil` in `get_pyr_tlt(self, p, c)` operates on plain Python scalars (`fft_sampling`/`fft_padding`, not xp arrays), so `math.ceil` would be simpler and avoid an unnecessary GPU round-trip on the cupy backend.

Menessao:
> * In `test_pyramid_tilt_coeffs`, the tilt is applied to face index 0, but the assertion checks the deviation of the last detected pupil (`[-1]`), and the naming looks swapped (`k[0,i]` multiplies `x`, but the test assigns `tlt_y` to it). This might well be correct if the focal-plane quadrant order doesn't map 1:1 to the detector pupil order, but could you confirm you checked this mapping directly rather than assumed it?
> 

Looking at the plt.imshow(frame,origin='lower') of the ccd frame, I believe the pupil ordering is as follows:
- _pupdata_calibrator_ : bottom left, bottom right, top right, top left
- _pyramid face tilts_ : top right, top left, bottom left, bottom right

Let me know if the updated tests make this more (or less) clear, or if any additional changes are required.

Regarding the x/y issue, I believe I simply had them swapped (twice) in the test, which should now be fixed

guidoagapito:
Thanks, the ceil fix, shape validation, and the all-ones regression test all look good.

One thing I would still like sorted out: I asked to Claude to look at `ids = [2,3,1,0]` against the mapping you described (pupdata: BL, BR, TR, TL / pyramid faces: TR, TL, BL, BR). It thinks that indices 0 and 1 match, but 2 and 3 look swapped — it gets `[2,3,0,1]`, not `[2,3,1,0]`.
The test only exercises `pupid=1`, which happens to be one of the two indices that is already correct either way, so this would not get caught as-is.
Could you loop the test over all 4 `pupid` values?
That settles it either way and gives good coverage going forward.

Also noticed `abs()` was added around the deviations in the assertions, so the test no longer checks the sign, only the magnitude.
Is it intentional?

The `pyr_tlt_coeff` type hint in the constructor signature is still float

```
                 ...
                 pup_shifts = (0.0, 0.0),
                 pyr_tlt_coeff: float = None,
                 pyr_edge_def_ld: float = 0.0,
                 ...
```

even though the `docstring` now correctly says list [2,4].
Could you update that too for consistency?

Menessao:
> Thanks, the ceil fix, shape validation, and the all-ones regression test all look good.
> 
> One thing I would still like sorted out: I asked to Claude to look at `ids = [2,3,1,0]` against the mapping you described (pupdata: BL, BR, TR, TL / pyramid faces: TR, TL, BL, BR). It thinks that indices 0 and 1 match, but 2 and 3 look swapped — it gets `[2,3,0,1]`, not `[2,3,1,0]`. The test only exercises `pupid=1`, which happens to be one of the two indices that is already correct either way, so this would not get caught as-is. Could you loop the test over all 4 `pupid` values? That settles it either way and gives good coverage going forward.
> 

I did manually loop over the 4 pupil ids and it was working, I have now added the loop in the test.

> Also noticed `abs()` was added around the deviations in the assertions, so the test no longer checks the sign, only the magnitude. Is it intentional?
> 

Yes, it was to avoid having to check signs. Do you expect the deviations to have the wrong sign? Should I check that as well?

guidoagapito:
Good, the test settles it empirically, which is more reliable than Claude's derivation from the written description anyway.

On the sign: fair enough, no need to change the test, it will be self-consistent either way.
It would be worth documenting what a positive coefficient actually does (which direction it pushes the pupil), though, so users know how to interpret the parameter without having to read the code.

Menessao:
> Good, the test settles it empirically, which is more reliable than Claude's derivation from the written description anyway.
> 
> On the sign: fair enough, no need to change the test, it will be self-consistent either way. It would be worth documenting what a positive coefficient actually does (which direction it pushes the pupil), though, so users know how to interpret the parameter without having to read the code.

Oh, I guess I should check that the given tilt coefficients are positive as well, otherwise the pupil ordering will get messed up.

Anyway, a coefficient greater than 1 pushes the pupil outward, away from the frame center, while a coefficient lower than 1 pulls it inward, towards the frame center, with respect to the nominal distance (given in input). 
The x coefficient (first row) shifts in the horizontal direction, the y coefficient (second row) shifts in the vertical direction.

I will check this tomorrow and clarify it in the doc string, if that is ok for you.

image
image

guidoagapito (approved):

PR #711, opened 2026-08-07
Powerloss computation update

bernadett92:
If doFresnel is set to true for the propagation the reference PSF is now computed using Fresnel propagation inside the PowerLoss class.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -41,6 +41,7 @@
 - Optimization of the compute\_ifs\_covmat function
 - Added Fraunhofer far field propagation
 - Fixed silent misparsing/confusing errors in split\_output() when an object, alias or output name contained a reserved '.', '-' or ':' character; added early validation of YAML section names in Simul
+- Updated calculation of power loss such that reference PSF also uses Fresnel propagation
 - ...
 
 ## [1.0.3] - 2026-05-18
--- a/specula/processing_objects/atmo_propagation.py
+++ b/specula/processing_objects/atmo_propagation.py
@@ -320,7 +320,7 @@ class AtmoPropagation(BaseProcessingObj):
 
         # pre-allocate arrays for propagation
         self.ef_padded = self.xp.zeros([self.ef_size_padded, self.ef_size_padded], dtype=self.complex_dtype)
-        
+
         # set wavelengthInNm field for all output EFs
         for source_name in self.source_dict.keys():
             self.outputs['out_'+source_name+'_ef'].wavelength_in_nm = self.wavelengthInNm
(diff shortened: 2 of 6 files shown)


guidoagapito:
Computing the reference PSF with the same Fresnel propagation as the real field makes the SR/power-loss comparison consistent, instead of comparing two things computed differently.

A few things to sort out before merging:

- In `prepare_trigger`, the reference field is propagated by calling `self.prop_obj.angular_spectrum_propagation(...)` / `fraunhofer_far_field_propagation(...)` directly on the live `AtmoPropagation` instance shared with the rest of the simulation.
Those methods write into `prop_obj.ef_fresnel`/`ft_ef1`, which are that object own internal scratch buffers, reused every real trigger.
It's probably harmless today (buffers get reset on the next real `trigger_code()`, and outputs are copied by value before `PowerLoss` runs), but it reaches into another object private state rather than a clean standalone API — fragile if `AtmoPropagation` internals change, and would become a real problem if processing objects are ever parallelized. Worth a standalone propagation helper that does not touch `prop_obj` own buffers.

- `self.prop_obj.common_layer_list[0].A` assumes there is always at least one common layer.
In practice that is almost always the `pupilstop`, so this will be fine in most configs, but `AtmoPropagation` itself only requires `atmo_layer_list + common_layer_list` combined to be non-empty, so a config with only `atmo_layer_list` set would hit a bare `IndexError` here instead of a clear message.
Worth a quick check with a proper error for that edge case.

- The reference field applies the pupil mask before all propagators, which matches the real propagation order only for `upwards=True` (the case this is tested with).
For `upwards=False`, `trigger_code()` reverses the layer order, so the reference would not match anymore.
Worth documenting this assumption explicitly, or raising a clear error if used with `upwards=False`, rather than leaving it implicit.

bernadett92:
I changed `angular_spectrum_propagation(...)` and ` fraunhofer_far_field_propagation(...)` such that they get a buffer as input. Then they do not require anything from `AtmoPropagation`, so I moved them outside the class.

guidoagapito:
Thanks, all three points are addressed well — moving the propagation functions out of the class is a cleaner fix than what I had in mind.

One more thing worth fixing: `self.ef_fresnel = angular_spectrum_propagation(self.ef_fresnel, propagator, self.ft_ef1, self.xp)` reassigns `ef_fresnel` to the returned buffer, which is `ft_ef1` itself — after the first call they become the same array instead of two independent buffers. It does not cause wrong results today, but it is relying on that incidentally rather than by design, and it is the kind of thing that could break silently if the function is ever touched again without realizing the two buffers have merged.

Cheap fix: write the final result back into `ef_in` instead of buffer, so buffer stays pure scratch and never gets aliased into the caller field:

```
def angular_spectrum_propagation(ef_in, propagator, buffer, xp):
    if propagator[0] is not None:
        ef_in *= propagator[0]
    buffer[:] = xp.fft.fft2(xp.fft.fftshift(ef_in, axes=(-2,-1)), axes=(-2,-1), norm="ortho")
    ef_in[:] = xp.fft.fftshift(xp.fft.ifft2(buffer * xp.fft.fftshift(propagator[1], axes=(-2,-1)), norm="ortho", axes=(-2,-1)), axes=(-2,-1))
    if propagator[2] is not None:
        ef_in[:] *= propagator[2]
    return ef_in
```

Same pattern for `fraunhofer_far_field_propagation`.
As a bonus, the caller no longer needs `self.ef_fresnel = ...`, just the call itself (closer to how it worked before the refactor).

Could you also add a small regression test asserting `prop.ef_fresnel` is not `prop.ft_ef1` after a Fresnel trigger?
The aliasing does not show up in the numeric results, only in object identity, so it is easy for it to silently come back later without a dedicated check.
The existing Fresnel propagation tests should already confirm the numeric output is unchanged, so this would just be to guard the identity specifically.

guidoagapito:
Another point: please update `ChangeLog.md`

guidoagapito (approved):

PR #713, opened 2026-08-17
Updated changeLog and fixed round/ceil issue in pyr mask computation

Menessao:
Opening as draft as there were merge conflicts for some reason


guidoagapito:
Makes sense — `round` matches the rounding convention already used elsewhere for these sizes, `ceil` did not for the odd-parity cases.
Good catch with the second test case.
Looks good, go ahead.

guidoagapito (approved):

PR #722, opened 2026-08-31
(GA) fast pseudo-inverse for IFunc

guidoagapito:


alfiopuglisi:
The test's comment mentions a singular value of "~2.6e-7 instead of numpy's ~2.8e-16". This looks like a problem with single/double precision confusion rather than a real issue, have you looked into this?

guidoagapito:
> The test's comment mentions a singular value of "~2.6e-7 instead of numpy's ~2.8e-16". This looks like a problem with single/double precision confusion rather than a real issue, have you looked into this?

You are right, this is single/double precision confusion, but on numpy side:
numpy.linalg.svd/pinv silently compute in float64 internally for any input dtype (see numpy.linalg.linalg._commonType, "in lite version, use higher precision"), then cast the result back down, regardless of BLAS backend (verified on reference BLAS, OpenBLAS, and numpy 2.1.1).
cupy cuSOLVER genuinely computes in float32, so its ~2.6e-7 residual is just the expected float32 noise floor.
Forcing a genuine float32 SVD on CPU (scipy with an explicit lapack_driver) reproduces both the ~1e-7 residual and the six-orders-of-magnitude pinv blowup — on CPU.
So this is not a GPU-specific limitation, it is a real float32-precision fragility of pinv on near-singular input that numpy default behavior happens to mask.
I fixed the docstring/comment accordingly.

alfiopuglisi (approved):

PR #726, opened 2026-09-03
Add window_xy for displays and Opticalib IM calib path

mbonaglia:
## Summary
- Add optional `window_xy` to `BaseDisplay` and pass-through on pixels/plot/slopec displays for lab window placement
- Register `OpticalibPushPullImCalibrator` in `CalibManager` under `im/`

## Test plan
- [ ] YAML display with `window_xy: [100, 200]` places the matplotlib window when the backend supports it
- [ ] Invalid/missing `window_xy` does not crash display setup
- [ ] SpecuLab Opticalib IM calibrator resolves data_dir under `im/` when used

### DIFF
--- a/specula/calib_manager.py
+++ b/specula/calib_manager.py
@@ -43,6 +43,7 @@ class CalibManager():
             'intmat': 'im/',
             'Intmat': 'im/',
             'ImCalibrator': 'im/',
+            'OpticalibPushPullImCalibrator': 'im/',
             'MultiImCalibrator': 'im/',
             'projmat': 'rec/',
             'RecCalibrator': 'rec/',
--- a/specula/processing_objects/integrator.py
+++ b/specula/processing_objects/integrator.py
@@ -1,4 +1,6 @@
 
+import numpy as np
+
 from specula.processing_objects.iir_filter import IirFilter
 from specula.base_processing_obj import InputDesc, OutputDesc
 from specula.base_value import BaseValue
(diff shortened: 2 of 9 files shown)


specula/calib_manager.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, root_dir):
             'intmat': 'im/',
             'Intmat': 'im/',
             'ImCalibrator': 'im/',
+            'OpticalibPushPullImCalibrator': 'im/',
```
alfiopuglisi, line 46: This line should be removed from the PR, this class is not inclded

specula/processing_objects/integrator.py
```diff
@@ -27,24 +29,32 @@ def __init__(self,
...
+            n_total = int(sum(n_modes))
             int_gain = [val for i, val in enumerate(int_gain) for _ in range(n_modes[i])]
             if ff is not None:
-                if isinstance(ff, int):
-                    ff = [ff]
-                if len(n_modes) != len(ff):
-                    raise ValueError(f"When n_modes is a list, length of n_modes {len(n_modes)}"
-                                     f" must match length of ff {len(ff)}")
-                ff = [val for i, val in enumerate(ff) for _ in range(n_modes[i])]
+                ff_arr = np.asarray(ff, dtype=float).ravel()
```
guidoagapito, line 46: These changes are not required/desired:
- support for scalar and expansion was already present
- we do not want to support truncation of longer ff vectors

alfiopuglisi: but then what is the actual desired change? The processing of "ff" looks identical to the one of "int_gain" in the original code, what is the feature to be added?

guidoagapito: I think the changes to this file must be rejected, no new feature is required.

PR #729, opened 2026-09-04
Fixing MPI bugs

alfiopuglisi:
Two MPI-related bugs fixed:
- trying to use a data object defined in the YAML file as an input only works locally, when connected via MPI it now raises an exception
- if one rank stops with an exception, the others could remain stuck forever, they are now aborted as well


alfiopuglisi:
Codecov is failing because the additional code is run using mpirun and is not detected by the coverage tool

specula/__init__.py
```diff
@@ -273,19 +274,28 @@ def main_simul(yml_files: list,
...
+                log_level=log_level,
+            ).run()
+    except Exception:
+        if mpi:
+            traceback.print_exc()
+            # Force other ranks to abort as well,
+            # otherwise they might remain stuck indefinitely
+            # waiting for input from this rank
+            MPI.COMM_WORLD.Abort(1)
+        raise
```
guidoagapito: A raise after Abort(1) is required?

FabioRossiArcetri (approved):

specula/__init__.py
alfiopuglisi: it's for the non-MPI case, I moved it inside an else: to be more explicit

guidoagapito (approved):

PR #730, opened 2026-09-15
Nn trainer tester

FabioRossiArcetri:


specula/data_objects/infinite_phase_screen.py
```diff
@@ -197,7 +244,28 @@ def get_new_line(self, row, after):
...
+        new_line = self.get_new_line(row, after)        
+        row_rms = self.xp.max(self.xp.abs(new_line))
+        new_line -= self.xp.mean(new_line)
+        new_line /= row_rms
+
+        if self.first:
+            self.lastmax = row_rms
+            self.first = False
+
+        #while row_rms > 5*self.lastmax:
```
alfiopuglisi, line 256: Do we need to keep the commented-out code?

requirements.txt
```diff
@@ -3,11 +3,10 @@ scipy
...
-astro-seeing >= 1.1
-synim == 1.2.3
+astro-seeing>=1.1
 flask-socketio
 python-socketio
 requests
 pytest
 scikit-image
-imageio-ffmpeg
+torch[neuralNetworks]
```
alfiopuglisi, line 12: This a very big dependency. Could it be an optional install like orthogram?

specula/processing_objects/atmo_infinite_evolution.py
```diff
@@ -1,107 +1,80 @@
...
         if not hasattr(extra_delta_time,"__len__"):
             self.extra_delta_time = cpuArray(self.n_infinite_phasescreens*[extra_delta_time])
         else:
             self.extra_delta_time = cpuArray(extra_delta_time)
 
+        self.psd1d_data = None
+        self.psd1d_freq_data = None
+        self.data_dir = data_dir
+        if psd1d_file:
+            self.psd1d_data = np.load(os.path.join(self.data_dir, psd1d_file))
```
alfiopuglisi, line 58: Is there a reason to use numpy files instead of FITS? it would be better to have a single storage format

alfiopuglisi (commented):
In addition to the inline comments, there is a lot of new code in this PR but no test coverage

